PR #37, opened 2020-03-31
Update text selection. Fix row index shifting.

### LINKED ISSUE #19: Mac: mouseDrag is raised even after a mouseUp
Not sure why, this causes the selection to be extended sometimes.

Repro:

in "bash$ foo" double click on the space between "$" and "foo", that should select that space, and quickly move a couple of lines down and quickly click - this currently extends the selection, because it receives a mouseDrag event, even when the mouse is up.

krzyzanowskim:
Most likely Fixes #19 

I changed a sequence a bit and don't rely on mouseDrag that heavily.

result:
selection 2020-03-31 22_05_54


migueldeicaza:
Woohoo!   Beautiful!  Thank you!

PR #53, opened 2020-04-08
Default monospacedSystemFont on Catalina. Use system font size

krzyzanowskim:
Default to `NSFont.monospacedSystemFont` on Catalina


migueldeicaza:
Oh this is wonderful!

PR #56, opened 2020-04-09
Use system colors (dark/light mode compatible)

krzyzanowskim:
Use system-defined colors for text/background/caret - this makes it work nicely in light/dark mode


migueldeicaza:
Oh another beautiful contribution, thanks for this!   It is a great touch!

PR #68, opened 2020-04-12
Use layout system to layout subviews. Subviews use bounds.

krzyzanowskim:
- Use `NSView.autoresizingMask` to layout and forget for subviews. Scroller and SelectionView
- Use bounds for subviews instead of frame.


migueldeicaza:
Beautiful, thank you!

Issue #71, opened 2020-04-13
Glitch when draw unicode character


krzyzanowskim:
Show this text repeatedly on the screen:

http://xn--4wa8awb4637h.org (Παν語 .org

![magic](https://user-images.githubusercontent.com/758033/79116148-fe604380-7d87-11ea-9c59-eec32de526f7.gif)

Observations: 
1. Lines below are not drawn properly
2. Window resize magically fixes the measurements shrug
3. Selection is broken - this is not the part of this bugreport as selection needs more work to share calculations with main view.

PR #73, opened 2020-04-13
Re-work text selection drawing

krzyzanowskim:
Re-work text selection drawing

- Delete `SelectionView` hack
- Selection is a text attribute and is drawn along the text.
    - draw selection color under text
    - (currently) overwrite text background color. It may (and should) be adjusted with some blending.

- Caret position is still broken for unicode characters https://github.com/migueldeicaza/SwiftTerm/issues/71, this will be addressed as a separate change


migueldeicaza:
Love it!   Thank you Marcin!

PR #77, opened 2020-04-16
useFixedSize uses fixed size

krzyzanowskim:
It looks to me that useFixedSize flag was reverted.


krzyzanowskim:
build error makes no sense to me

Screenshot 2020-04-16 at 02 10 07

migueldeicaza:
Oh you are right - the problem was the boolean declaration above was inverted.   Your patch changes a number of other things, so I merged it rather than changing the two lines on my end.

migueldeicaza:
The breakage was the missing file.

PR #80, opened 2020-04-16
Rework glyphs drawing. Use fixed font size measurements.

### LINKED ISSUE #67: Combining characters
I suspect that the reason the UTF-8-Demo.txt regressed in spacing with the math symbols is my new approach at combining characters.

krzyzanowskim:
Fixes #67

- [x] revert to fixed height/width for size, yet still draw glyphs that cross the lines
- [x] disable selection view (again) and fixes performance issue when updating selection
- [x] update caret view position along the lines
- [x] draw underline
- strikethrough doesn't (never did)


SwiftTerm/Sources/SwiftTerm/Mac/MacTerminalView.swift
```diff
@@ -115,6 +114,9 @@ public class TerminalView: NSView, NSTextInputClient, NSUserInterfaceValidations
         
     func setup(frame: CGRect, bounds: CGRect)
     {
+        wantsLayer = true
```
migueldeicaza, line 117: What do we use the layer for?   Is this a good idea?

migueldeicaza:
This looks very good, let me test it a little bit locally tonight.

SwiftTerm/Sources/SwiftTerm/Mac/MacTerminalView.swift
krzyzanowskim: This is a good question.

I used it to have a background without drawing because when drawing in context, it interferes with how selection/background color blending.

I didn't notice a side effect, but now I'm not sure

on one hand:

https://developer.apple.com/documentation/appkit/nsview/1483695-wantslayer
> In a layer-backed view, any drawing done by the view is cached to the underlying layer object. This cached content can then be manipulated in ways that are more performant than redrawing the view contents explicitly. 

on the other

https://developer.apple.com/documentation/appkit/nsview/1483686-drawrect
> If your app manages content using its layer object instead, use the updateLayer method to update your layer instead of overriding this method.

krzyzanowskim: I think we're good, after checking wantsUpdateLayer()

https://developer.apple.com/documentation/appkit/nsview/1483461-wantsupdatelayer

> A view can update its contents using one of two techniques. It can draw those contents using its draw(_:) method or it can modify its underlying layer object directly. During the view update cycle, each dirty view calls this method on itself to determine which technique to use. The default implementation of this method returns false, which causes the view to use its draw(_:) method.

migueldeicaza:
This imrproves the rendering performance and fixes selection.

It makes the simple benchmark go from 0m0.827s to  0m0.538s.

migueldeicaza:
If we can turn off the layer, we could turn it off.   Might be more resource friendly.

migueldeicaza:
Well `wantsLayer` at least renders the bits below the screen, so I guess that is fine?

That said, I just added code to render under there, and that should have been triggering

migueldeicaza:
With this new patch, this looks glorious:

image

PR #82, opened 2020-04-18
Don't invoke the superclass of mouse responder methods

krzyzanowskim:
https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/EventHandlingBasics/EventHandlingBasics.html

> If your implementation of an NSResponder method such as mouseDown: completely handles an event it **should not invoke the superclass** implementation of that method.


migueldeicaza:
Thank you!

PR #83, opened 2020-04-19
terminal view options

krzyzanowskim:
Refactor terminal view options into single property: `options`. This property is intended to be modified by the user, and the view should apply the changes.

Put this snippet in TerminalView.setup() and see how it change colors:

```swift
DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(5)) {
    let o = Options(font: Options.default.font, colors: Options.Colors(useSystemColors: true))
    self.options = o
}
```

color-change

### DIFF
--- a/SwiftTerm/Sources/SwiftTerm/Terminal.swift
+++ b/SwiftTerm/Sources/SwiftTerm/Terminal.swift
@@ -366,10 +366,10 @@ open class Terminal {
         // call this
         cols = max (options.cols, MINIMUM_COLS)
         rows = max (options.rows, MINIMUM_ROWS)
-        if isReset {
+        if buffers != nil && isReset {
             buffers.resetNormal ()
             buffers.activateNormalBuffer(clearAlt: false)
-        } else {
+        } else if buffers == nil {
             buffers = BufferSet(self)
         }
         cursorHidden = false
--- a/SwiftTerm/Sources/SwiftTerm/Terminal.swift
+++ b/SwiftTerm/Sources/SwiftTerm/Terminal.swift
@@ -152,7 +152,7 @@ open class Terminal {
     /// The current terminal rows (counting from 1)
     public private(set) var rows : Int = 25
     var tabStopWidth : Int = 8
-    var options: TerminalOptions = TerminalOptions()
+    var options: TerminalOptions
     
     // The current buffers
     var buffers : BufferSet!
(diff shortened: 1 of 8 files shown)


SwiftTerm/Sources/SwiftTerm/Terminal.swift
```diff
@@ -366,10 +366,10 @@ open class Terminal {
         // call this
         cols = max (options.cols, MINIMUM_COLS)
         rows = max (options.rows, MINIMUM_ROWS)
-        if isReset {
+        if buffers != nil && isReset {
```
migueldeicaza, line 369: Curious about the reason for this change.

krzyzanowskim: `buffers` is "forced unwrapped", so it's like Optional. Here, for sake of check, buffers can't be `nil`.

The change here says: do not touch (clear) buffers unless `isReset` is set.

krzyzanowskim:
> a new set of options that only apply to the Mac View, and I assume those would allow changes dynamically?

yes. `TerminalView.Options` apply to the view only, hence `TerminalOption` is not exposed. We probably should expose TerminalOptions too, however, because of how it depends on the view, I left it behind.

migueldeicaza:
I attempted to fix the conflict.

migueldeicaza:
Mhm, I wonder if we should have a getter/setter API that controls those values?   Setting all in one go does not seem like a good idea.

krzyzanowskim:
>  Setting all in one go does not seem like a good idea.

I thought about this kind of API, and it was designed after `TerminalOptions`

```swift
var options = TerminalView.Options(........)
terminalView.options = options

// now change something
options.fonts.normal = NSFont.myfont
options.colors.foregroundColor = .black
terminalView.options = options
```

this approach assumes that `TerminalView` internally resolve dependencies and apply detected changes in one go

alternate API is:

```swift
terminalView.font = NSFont.myfont
terminalView.foregroundColor = .black
```

this is more AppKit API. Internally we still have to solve the order of the updates

migueldeicaza:
I see - well, `TerminalOptions` is a little bit of a hack, I have never liked it, so I think we are safe changing the style - since you did all the work that actually does all the heavy lifting, I think we could change to be more UIKit/AppKit like.

I can do it if you want, just wanted to make sure you were ok with it.

PR #94, opened 2020-04-25
Move SPM manifest to the root so SPM can use it now

krzyzanowskim:
Fix SPM support


migueldeicaza:
I merged, but not sure what this is for :-)

migueldeicaza:
Oh now I realize why - so that third parties can easily use it.   Thanks!

PR #96, opened 2020-04-25
add terminator to byte array that is read a null terminated c string

palmin:
I was running with memory bounds checks and found this


migueldeicaza:
Thank you!   The one place I use unsafe code, is the place where the bug is.

I hate that I need to go there to create a String, I wish I could create it out of the array directly.

Issue #93, opened 2020-04-25
Should review API to surface suitable APIs for screen-scraping


migueldeicaza:
Now that I made HeadlessTerminal public, as I need it for fuzzing, perhaps I should surface more information about the contents of the screen and make a handful of internal classes public.

PR #102, opened 2020-04-29
Expose Terminal information outside module

palmin:
A few changes to allow using the Terminal object from outside the SwiftTerm module.

You can read the following:
* CharData for a given row and column
* whether terminal is in application cursor mode or not
* cursor location relative to visible screen
* top visible row to help handle scrolling


migueldeicaza:
Brilliant!   Thank you!  

This aligns beautifully with this: https://github.com/migueldeicaza/SwiftTerm/issues/93

PR #110, opened 2020-05-02
conversion of sixel graphics into images 

palmin:
I was satisfied with the decoding of sixel formatted images but I keep stumbling on the best way to keep track of and render these images after decoding.

Perhaps this pull request that only does decoding is useful. 

It registers a DcsHandler and delivers a CGImage to the new `Terminal.sixel (_ image: TTImage)` function that doesn't yet do anything.


migueldeicaza:
Thank you, this is fabulous!   I am going to check it out tomorrow morning!

migueldeicaza:

PR #135, opened 2020-07-31
Removed an unused local variable buffer that blocked compilation

pmkary:


migueldeicaza:
I don't mind the change, but why does it block compilation?

migueldeicaza:
Used a different approach, as this was introduced during some refactoring, but lost the performance benefit - this showed up again on the profiling.

PR #148, opened 2020-11-08
fix true color display

lvht:


migueldeicaza:
Can you explain this change?

The colors are passed as 16-bit values, and I am wondering why you end up with 8 bit values.

lvht:
I use the master branch, and it does not display ture color at all. It seems the r,g,b has a range of  0-255

lvht:
The class Color's  red/green/blue do have a range of 0-65535. And the Attribute.trueColor is defined as
```swift
        /// This means that the color has been configured to be a 24-bit true color
        /// and has 8 bits for red, green and blue
        case trueColor(red: UInt8, green: UInt8, blue: UInt8)
```

lvht:
BTW, this is an awesome project to use and to learn. Would you like move SwiftTerm to a dedicate project? It seems that the github.com/SwitfTerm is available.

migueldeicaza:
Can you provide me with the test case?

Will review later today!

And no, I do not think I want to move this to a separate organization.

migueldeicaza:
@lvht I do not appreciate the creation of a SwiftTerm organization, please delete it.

lvht:
I will do independent on the new organization.

migueldeicaza:
@lvht that is really poor form to fork someone else's project and hijack the name.

migueldeicaza:
I finally had time to review this, it is indeed 0..255, applied your fix.

PR #154, opened 2020-12-02
avoid buffer overflow in selectCharset

palmin:
`selectCharset()` could overflow when called with incorrect parameters.

`if CharSets.all.keys.contains(p [1]){` a few lines down fails if `count < 2` and by returning early this can no longer happen.


migueldeicaza:
Thank you for this change!

Issue #133, opened 2020-07-14
iOS: implement UIScrollView support


migueldeicaza:
Currently the iOS view lacks UIScrollView support.

migueldeicaza:
Work in progress in the `uiscrollview-hacks` branch, as I put some ugly hacks in place

migueldeicaza:
Update, August 2022:

New branch `uiscrollview-2022-07` has a cleaner way of supporting UIScrollView.

Current issues in the branch:

- Need to update the cursor position, as it is not aware of this new approach at being positioned in the screen
- Currently the UIScrollView conflicts with the UIPanGestureRecognizer that I define for mouse input.   There are two ways of solving this problem: 
  - adding my own Pan gesture recognizer only when the capabilities that need it require it (allowMouseReporting && terminal.mouseMode != .off is one, the other one is selection, which I think might be worth only turning on if there is a criteria for doing so).   
  - Using the built-in system that allows two gesture recognizers to work together, but it might be tricky or ugly to get this done, and would result in more obscure core.
- I probably should set the scrolling size in app mode to be fixed, and disable UIScrollView entirely in that mode.
- There is a bug where leaving "mc" leaves some of that garbage behind, not sure why.
- `ls` sometimes shows a handful of lines missing at the start for a long scroll, no idea why.
- Need to automatically scroll to the bottom on keyboard input

migueldeicaza:
This has now been merged and all issues addressed

Issue #99, opened 2020-04-27
iOS App Features


migueldeicaza:
- [ ] Sort out scrolling (UIScrollview?)
- [ ] Double tap, becomes tab
- [ ] Map touches to mouse events
- [ ] Terminal-bar (escape, control, alt, settings)
- [ ] Swipe gestures to send up/down/left/right

palmin:
I'm experimenting with putting in a layer between Terminal and TerminalView for producing NSAttributtedString in a way that can be used in UITextView which would support scrolling and text selection.

Not sure if this is a good idea but seems worth a try.

kirb:
I’d like to port my app [NewTerm](https://github.com/hbang/NewTerm) over to SwiftTerm. It’s a jailbreak terminal app, originally based on MobileTerminal (which is almost as old as the iPhone itself!) and using a fork of iTerm (not iTerm2) as the emulator. I’ve planned to shift away from the ancient and broken emulator, and give it an SSH backend for release on the App Store. Porting xterm.js to Swift was a project I started working on, but you’ve definitely gotten way further than me on that 😁

Is there anything of interest that I could work on or provide here? I have some ideas on some UI aspects. I’ll be relicensing NewTerm under Apache License 2.0 once all GPL iTerm code has been removed. No original MobileTerminal code remains anywhere but in the Common/VT100/ directory.

krzyzanowskim:
> I'm experimenting with putting in a layer between Terminal and TerminalView for producing NSAttributtedString in a way that can be used in UITextView which would support scrolling and text selection.

To make it 1:1 we'd need to move from CoreText to TextKit (TextKit API is pretty annoying in comparison). There's a couple of problems to solve on the way.

migueldeicaza:
Apologies, I only saw this comment now.

On scrolling: I had some early bits in place, I committed them now, where I made the terminal view be a subclass to UIScrollView - I am not attached to this model, we can try other approaches.

I think this is a good match, but I need to think further.   The reason why I think this works is because the scroll buffer grows, and I would not want to render the entire back-scroll, I would like to render the back scroll on demand.

I do not know much about UITextView as a backing engine for this, will need to research what this means. 

For Sixel graphics, I think that we could flag a cell as having an image, and then the rendering code that is in place today would paint the image instead of adding this to the CTRun.

@kirb: In terms of what you could work on, any of the above gestures would work.    I collected some of those from using terminals in the past and felt like these would be useful.

[...]

PR #156, opened 2020-12-08
Hack for scrolling

skyline75489:
I'm not sure this is the correct way to do this but here it is.

Currently UIScrollView does not work. And I don't think there is a possible way for it to work. The gestureRecognizers added will conflict with those inside UIScrollView, making UIScrollView totally unusable.

So I just make it plain UIView and put together scrolling logic. This way there'll be no fancy scrollbar or bouncing, but the scrolling will work.

### DIFF
--- a/Sources/SwiftTerm/iOS/iOSTerminalView.swift
+++ b/Sources/SwiftTerm/iOS/iOSTerminalView.swift
@@ -32,7 +32,7 @@ import CoreGraphics
  * Use the `configureNativeColors()` to set the defaults colors for the view to match the OS
  * defaults, otherwise, this uses its own set of defaults colors.
  */
-open class TerminalView: UIScrollView, UITextInputTraits, UIKeyInput, UIScrollViewDelegate {
+open class TerminalView: UIView, UITextInputTraits, UIKeyInput {
     struct FontSet {
         public let normal: UIFont
         let bold: UIFont
@@ -267,7 +267,22 @@ open class TerminalView: UIScrollView, UITextInputTraits, UIKeyInput, UIScrollVi
             // endEditing(true)
         }
     }
-    
+
+    private func calcScrollingVelocity (delta: Int) -> Int
+    {
+        if delta > 9 {
+            return max (terminal.rows, 20)
+        }
+        if delta > 5 {
+            return 10
+        }
+        if delta > 1 {
+            return 3
+        }
+        return 1
+    }
+
+
     @objc func pan (_ gestureRecognizer: UIPanGestureRecognizer)
     {
         guard gestureRecognizer.view != nil else { return }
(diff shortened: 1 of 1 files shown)


migueldeicaza:
I have a branch where I did some of the work, but it has still a few bugs here and there.

https://github.com/migueldeicaza/SwiftTerm/issues/133

I agree the gesture recognizer will have to be toggled on/off under this scenario, I was tracking that here (last bullet point at the time of this writing:

https://github.com/migueldeicaza/SwiftTerm/issues/99

So I want to have a toggle on/off for the scroll events, so users can control whether they want the gestures to work, or the scrolling to work.

How are you using this scrolling if you are setting up the UIView directly?   Your approach might be better than what I did.   Or how are you testing it?

skyline75489:
I'm basically taking the same approach as on macOS, to manually handling scrollUp/scrollDown based on Pan gestures. Because after some experiments, I think that UIScrollView is just not suitable for a terminal app like this. 

Due to the nature of UIScrollView, whatever `drawRect` added to the context will behave like a static picture in the background during the scrolling. UIScrollView handles all the gestures in its own implementation and does not allow us to intercept it and customize it. Ideally we want to intercept the scrolling gesture and transform it into scrollUp/Down for the terminal and then call `setNeedDisplay()` to refresh it. But it's just not possible. UIScrollView will always scroll things (presumably subviews) by transforming the origin. 

If UIKit has `NSScroller` as in AppKit then we should also be able to add a scroller on the right side. But that's not the case. So I had to ditch UIScrollView entirely and just use a plain UIView. Sure it's not as smooth as UIScrollView, but it gives us more control over the behaviour of gesture handling.

As for the mouse handling on iOS, I haven't really figured it out yet. I used pan gestures for scrolling and it will conflict with the mouse events, as you can see in the code. I feel that mouse support is not as important as scrolling. Without scrolling, the terminal is not usable. Without mouse support the terminal will still be usable to most people. Besides with iOS being able to use trackpad & mouse now, handling mouse events correctly is even more harder. If it's OK I think the mouse support can be left for future work.

I'm only a person who is exploring terminals on iOS devices. Feel free to close this PR if you don't feel comfortable about it.

migueldeicaza:
Hello!

I have implemented scrolling support now, and sorted out the conflicts between sending events vs moving around.   It is a setup that works, but does not feel very iOS-ish.

It does not use the UIScrollView capability, like you point out, and that still makes me sad.

I have some comments from some months ago here: https://github.com/migueldeicaza/SwiftTerm/issues/133 

The code changed since I had that branch, and I think I have a different idea of how I want to limit the rendering.    I will try again that branch, which made more progress than my ill-attempted effort last night (last night I tried a quick and dirty "What if I move the terminal to a separate view", which was too dirty to work;   but also found that gestures were not working, despite disabling them).

The research from last night was interesting because it did surface that there is a series of methods that UIScrollView can use to interoperate with other gesture handlers.   So I will try to learn more about those ("Managing Touches" and the intro text in the documentation here: https://developer.apple.com/documentation/uikit/uiscrollview)

migueldeicaza:
Mhm, I do like your change for scroll velocity, I will rebase tonight and try it out.

Thanks!

skyline75489:
@migueldeicaza happy to help 😃 

Btw, thanks for defending the terminal devs (me included or not, doesn’t matter) on Twitter about a week ago. That was very kind of you to take your time and speak out.

migueldeicaza:
Ok, I have implemented this in the new branch: uiscrollview-2022-07, doing some touchups before I merge

PR #160, opened 2020-12-10
Make ctrl-6 works for vim

lvht:


migueldeicaza:
Can you explain the reason for this change?   On the MacOS Terminal.App control-6 does not do anything, except beep that this is an incomplete input.

lvht:
If you use vim, use could see the doc for `CTRL-6` by typing `:h CTRL-6`. And it reads

> If the 'autowrite' or 'autowriteall' option is on and
the buffer was changed, write it.
Mostly the ^ character is positioned on the 6 key,
pressing CTRL and 6 then gets you what we call CTRL-^.
But on some non-US keyboards CTRL-^ is produced in
another way.

In iTerm2, pressing CTRL and 6 is equal to CTRL and ^, by which I can let vim edit the alternate file.

migueldeicaza:
You are right, iTerm2 supports this.

PR #165, opened 2021-01-23
Stretch last col/row to fill the view frame

krzyzanowskim:
Stretch last col/row to fill the view frame.

Problem: If the view frame is not exactly the size of the grid, the background does not cover the whole frame - that looks like here:

https://user-images.githubusercontent.com/758033/105563386-2a98ef80-5d1e-11eb-8dd0-ed37ff5e7004.mov


migueldeicaza:
Thank you for tracking and fixing this bug!

PR #168, opened 2021-04-04
Secure ShellFish changes

palmin:
The work in these commits is covered by the MIT license. 

After making a similar pull request #138 I unfortunately never merged back from upstream and just kept working on my fork. This makes this pull request messier than it could have been and cherry picking commits might be the least-bad approach.

The most significant change is [explicit close](https://github.com/migueldeicaza/SwiftTerm/commit/59bbb505da1d74d54c8d94280475d28fa08465a5) that breaks up retain loops inside Terminal to allow memory to be freed when a Terminal instance is no longer needed.

I have also done work to support [tmux control mode](https://github.com/tmux/tmux/wiki/Control-Mode) inside EscapeSequenceParser but it wasn't completed yet and I removed it while preparing this PR as it caused a lot of merge conflicts and I wasn't happy with the approach. 

It can be found inside the commits included in this PR even though it didn't make it into the head commit. I hope to implement a nicer approach to tmux control mode in the next months.

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -13,6 +13,8 @@ jobs:
   build:
 
     runs-on: macos-latest
+    env:
+      DEVELOPER_DIR: /Applications/Xcode_12.2.app/Contents/Developer
 
     steps:
     - uses: actions/checkout@v2
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "SwiftSH.binaries"]
-	path = SwiftSH.binaries
-	url = git@github.com:migueldeicaza/SwiftSH.binaries.git
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ this module.
 
 Check the [API Documentation](https://migueldeicaza.github.io/SwiftTerm/)
 
-The macOS AppKit NSView implemention [`TerminalView`](https://migueldeicaza.github.io/SwiftTerm/Classes/TerminalView.html) is a reusable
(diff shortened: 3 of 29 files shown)


migueldeicaza:
Thank you for sending along the PR.

I was going to start pulling few bits as you suggested as PRs, and then I realized that maybe I want to take a look at fixing the ownership problem in the terminal using weak properties, as the various connected classes (parser, bufferset, buffer) are keeping the terminal alive, but can only ever be alive, if the container is alive - which sounds like a perfect use for weak.

I have a partial patch here, and while I believe it works, I want to validate that it works with Instruments.   I have had very little luck with Instruments and the Leak detection with a standalone program.   It does not show up as a leak, but I can see it stick around, even with my fix, so I will investigate if there are other options I can take to automatically dispose that.

https://gist.github.com/migueldeicaza/59e53ce8109784b4396b24b03622810b

migueldeicaza:
Ok, this looks very promising, it shows that it works, and does not require the user to manually call `close` on the terminal:

https://gist.github.com/migueldeicaza/5b84c01c5695ef1bbd08a74fb2d27675

migueldeicaza:
Thank you!   I think I got everything in place manually.

The only differences are as follows:

(a) I ended up fixing the internals, so `close` should not be necessary.

(b) the line changed callback should not be necessary, as there is now a set that tracks those changes, and is exposed.

palmin:
Looks very nice. 

I will be switching Secure ShellFish to your official version of SwiftTerm as that should make it easier to incorporate changes from me going forward.

PR #199, opened 2021-11-05
LocalProcessTerminalView can reuse an already started process

klein-artur:
This is an idea how to solve two different issues:

1. You can close one terminal, and reopen a new one with the same process. The way to do this would first getting the process from the current `localProcessTerminalView.process` and save it. Then you can create a new `LocalProcessTerminalView` by calling the initializer with the saved process `LocalProcessTerminalView(frame: frame, localProcess: savedProcess)`.
That way the terminal will open in the state of the process, not in a new state.

2. I personally used this to make my terminal window resizable. What I do is, in my draw method of the containing view I create a new terminal, with the new frame but with the old process. That way the terminal will adapt the size perfectly.


klein-artur:
wrong pull request, sorry.

PR #200, opened 2021-11-15
Use serial queue to send bytes

krzyzanowskim:
This looks like a typo, documentation says the `dispatchQueue` will be used to post data.
The handler code is not thread-safe anyway.

### DIFF
--- a/Sources/SwiftTerm/LocalProcess.swift
+++ b/Sources/SwiftTerm/LocalProcess.swift
@@ -97,7 +97,7 @@ public class LocalProcess {
                 print ("[SEND-\(copy)] Queuing data to client: \(data) ")
             }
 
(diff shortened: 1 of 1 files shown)


migueldeicaza:
The `dispatchQueue` is the queue used to send data from the local process back to SwiftTerm, that is because access to the internal sate of SwiftTerm must be done from the thread controlling it.

But this queue is not about parsing the data being received, this is a queue used to send the data to the child process.

Can you explain what you mean by `The handler code is not thread-safe anyway.`?

krzyzanowskim:
> Can you explain what you mean by The handler code is not thread-safe anyway.?

the `total` var is updated from the concurrent queue, possibly from multiple threads at a time, however the access is not synchronized.

migueldeicaza:
I do not think that this is the fix for the problem - see the comments here:

https://github.com/migueldeicaza/SwiftTerm/issues/201

PR #204, opened 2022-01-12
Terminate local process

krzyzanowskim:
Adds a convenient function to terminate the child process

### DIFF
--- a/Sources/SwiftTerm/LocalProcess.swift
+++ b/Sources/SwiftTerm/LocalProcess.swift
@@ -53,7 +53,7 @@ public class LocalProcess {
     public private(set) var childfd: Int32 = -1
     
     /* The PID of our subprocess */
-    var shellPid: pid_t = 0
+    public private(set) var shellPid: pid_t = 0
     var debugIO = false
     
     /* number of sent requests */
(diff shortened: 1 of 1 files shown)


migueldeicaza:
I do not mind surfacing the shellPid, but the policy implemented in `terminate()` seems arbitrary and I have two concerns.

The use of waitpid can be problematic, and I need to do some research:

`waitpid` behavior can be affected by a global `sigaction (SIGCHLD, SIG_IGN)` or if `SA_NOCLDWAIT` has been set for `SIGCHLD`) - yet, I do not know what could be done to properly track the state of `shellPid` if I were to just make it public.

Sadly, the `sigaction` state for `SIGCHLD` is a global process setting.

The loop is messy, because if we are giving a chance to the process to terminate, and the process has not finished, sending another signal might just trigger the same cleanup processes again.   So I would rather just have a larger delay than then loop.

Perhaps an option is `terminate (staged:Bool=true, waitForChild:Bool=true)` to give folks flexibility if they want to use the staged kill, or not, and whether we should call `waitpid`.

krzyzanowskim:
ok, maybe just `kill(shellPid, SIGTERM)` and hope it will die. I tried to be smart, but maybe there's no point in it. That'd be in line with Swift `Foundation.Process` https://github.com/apple/swift-corelibs-foundation/blob/efedfc30e7ea2ce5e4db85e3869b9cb149273ad9/Sources/Foundation/Process.swift#L1032

Co2333:
> The use of waitpid can be problematic, and I need to do some research:

You might want to have a try on dispatch source, I have used it [here](
https://github.com/Co2333/AuxiliaryExecute/blob/main/Sources/AuxiliaryExecute/AuxiliaryExecute%2BSpawn.swift#L269) and it works like a charm.

migueldeicaza:
@Co2333 Funny bit, I was already using DispatchSource to catch the child signal, so I guess we do rely on a swift-installed sigaction and swift-global sigaction semantics.   

I think I will keep the `terminate()` from Marcin, and also surface the child PID for folks that might want to use it.

Issue #227, opened 2022-04-17
Chinese characters are chopped


migueldeicaza:
Given a chinese character that uses more than one cell, and another one, we get two CTRuns, one for the Chinese character, and one that contains both a space and the second character.   The space when we render seems to be overwriting the second half of the Chinese character.

The patch below extracts the position of the rune relative to the line, and helps a bit, but now the last character of the line is chopped, so there is some clipping somewhere I need to find.

```
diff --git a/Sources/SwiftTerm/Apple/AppleTerminalView.swift b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
index 59f8643..4ba20e0 100644
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -564,9 +564,13 @@ extension TerminalView {
                     CTRunGetGlyphs(run, CFRange(), bufferPointer.baseAddress!)
                     count = runGlyphsCount
                 }
+                let x = [CGPoint](unsafeUninitializedCapacity: runGlyphsCount) { (bufferPointer, count) in
+                    CTRunGetPositions(run, CFRange(), bufferPointer.baseAddress!)
+                    count = runGlyphsCount
+                }
 
                 var positions = runGlyphs.enumerated().map { (i: Int, glyph: CGGlyph) -> CGPoint in
-                    CGPoint(x: lineOrigin.x + (cellDimension.width * CGFloat(col + i)), y: lineOrigin.y + ceil(lineLeading + lineDescent))
+                    CGPoint(x: lineOrigin.x + x [i].x /* (cellDimension.width * CGFloat(col + i))*/, y: lineOrigin.y + ceil(lineLeading + lineDescent))
                 }
 
                 var backgroundColor: TTColor?
```

JoeMatt:
I think you need to convert the whole string to a width, not just the glyphs, especially in Unicode which can have reverse text, subscripts etc.

andy380743909:
> I think you need to convert the whole string to a width, not just the glyphs, especially in Unicode which can have reverse text, subscripts etc.
> 
> 
> 
> 
> 
> https://developer.apple.com/documentation/foundation/nsstring/1524729-boundingrectwithsize

I think it's not the string width but glyph postions in line are wrong. You just point out a right direction of this issue, because the CoreText calculate each character's position not just using the char width multiply the index of it in a line.
[...]

PR #239, opened 2022-06-10
[bugfix] Chinese characters are chopped #227

andy380743909:
https://github.com/migueldeicaza/SwiftTerm/issues/227

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -572,10 +572,34 @@ extension TerminalView {
                     count = runGlyphsCount
                 }
 
+                #if false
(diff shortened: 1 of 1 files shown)


andy380743909:
sorry, this PR does not totally fix the issue. because the space char after every Chinese char engage a cell.
I'm not very clear with the logic of the space char. Can I just remove it?

migueldeicaza:
Thanks for the patch, I think this is a step in the right direction, but I need to rethink a bit this code.

andy380743909:
> Thanks for the patch, I think this is a step in the right direction, but I need to rethink a bit this code.

yes, I agree with you. And I read the iTerm2 source code, but it's really too complex, I'm searching for the NSAttributedString building logic in it's source code.

https://github.com/gnachman/iTerm2/blob/1a0c9a53616abfdb028a67b5a62dc6e34af6fc59/sources/iTermTextDrawingHelper.m#L9

PR #242, opened 2022-06-30
Add default value to `updateCursorStyle()` in `MacCareView.swift`

JoeMatt:
Fixes build issue in `AppleTerminalView.swift` line 64
```swift
    func updateCaretView ()
    {
        caretView.frame.size = CGSize(width: cellDimension.width, height: cellDimension.height)
        caretView.updateCursorStyle()
    }
```


migueldeicaza:
Thanks for the PR!   I ended up doing a slightly different change.

PR #268, opened 2023-01-05
[bug fix]  fix incidental double tap on up arrow

iOSappssolutions:
https://github.com/xibbon/LaTerminal/issues/19

I think this solves issue 19


migueldeicaza:
Thank you for the fix!

PR #273, opened 2023-02-04
Fix delegate

krzyzanowskim:
Fix building MacLocalTerminalView.swift


migueldeicaza:
Thank you Marcin!   Apologies for not catching this one!

PR #287, opened 2023-03-29
fix space on bottom of terminal view without color

iOSappssolutions:


migueldeicaza:
Thank you for the patch!

PR #291, opened 2023-04-18
This change only queues the exposed region to be repainted on UIKit.

migueldeicaza:
This works, but it exposes a problem: scrolling of text (ls -l on a large directory) will redraw the text on top of the old text - this is caused by the
clear background that I allow using.   So I need to account for that.

This sadly also brings back the ugly bar at the bottom, because the UIScrollView allows for over-scroll, and I do not seem to have a way of painting there.  The solutions on StackOverflow include "Append a view underneat with the color that
you want" which is some ugly stuff.   Will investigate later.


migueldeicaza:
I fixed the repaint issues on main, so this patch can likely go in soon.

migueldeicaza:
Too old, should some day.

PR #293, opened 2023-05-07
Use weak ref to fix crash when process exits after view is released

kdrag0n:
If a LocalProcess is terminated or exits right before the hosting LocalProcessTerminalView is released, the processTerminated call crashes the app with the following error:

```
Fatal error: Attempted to read an unowned reference but the object was already deallocated
```

All usages of delegate already have nil checks, so just change the variable from unowned to weak to fix the issue. There's no point in calling the delegate when it no longer exists.

### DIFF
--- a/Sources/SwiftTerm/LocalProcess.swift
+++ b/Sources/SwiftTerm/LocalProcess.swift
@@ -68,7 +68,7 @@ public class LocalProcess {
     var sendCount = 0
     var total = 0
 
-    unowned var delegate: LocalProcessDelegate?
+    weak var delegate: LocalProcessDelegate?
     
     // Queue used to send the data received from the local process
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Thank you for your patch!

migueldeicaza:
Mhm, I wonder why changing unowned to weak fixes this?

kdrag0n:
I think it's because `unowned` is strict and crashes if you try to access a reference to a freed object. `weak` returns `nil` instead, which is more in line with the expected behavior for delegates.

Lakr233:
`unowned` should be banned for 99% of cases, it is dangerous as we do not have, most of time, control to ARC overtime so the pointer should be set to nil when release (`weak`) instead of keep it (`unowned`)

PR #310, opened 2023-06-18
Fix CI

ichizok:
And I want to confirm:

* Should disable pages-build-deployment action; it always fails since there is no longer "docs" directory.
* Why use your mirror esctest instead of the original repo for testing?

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -7,38 +7,34 @@ on:
     branches: [ main ]
 
 env:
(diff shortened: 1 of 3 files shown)


migueldeicaza:
Thanks!  I want to move to DocC for documentation, and follow what I did in SwiftGodot: which is to push the generated docs to a separate tree (because otherwise it makes it hard to search on the repo)

The esctest is so I can control which version is validated against and in case I make adjustments on the test suite.

PR #321, opened 2023-08-12
Use if canImport(...) instead of if os(...)

migueldeicaza:


Lakr233:
iirc, this will result a failure in Mac Catalyst app which can both import AppKit and UIKit. correct me if I'm wrong. :P

截屏2023-08-12 21 05 14

截屏2023-08-12 21 07 38

migueldeicaza:
Thanks for the review.

I have not been considering catalyst, but that is a good call.   I am afraid that generally the codebase is not really ready for Catalyst in this scenario :-(

Lakr233:
> Thanks for the review.
> 
> I have not been considering catalyst, but that is a good call. I am afraid that generally the codebase is not really ready for Catalyst in this scenario :-(

Actually me and my team port your SwiftTerm to tvOS for fun and it works like a chram. :P

Issue #180, opened 2021-05-21
Extended Keyboard Input Support


migueldeicaza-msft:
There is a proposal floating around to extend the keyboard support in terminals:

https://sw.kovidgoyal.net/kitty/keyboard-protocol.html

Might be worth adding.

migueldeicaza:
Work in progress branch:

https://github.com/migueldeicaza/SwiftTerm/compare/main...extended-keyinput

migueldeicaza:
This is how tmux implemented it:

0ab82d95314e7a26a48452c77ad710f3aff97dd7

migueldeicaza:
Updated version: `kitty-input`

migueldeicaza:
Finished.

PR #323, opened 2023-09-03
fix: control combos without special mappings are eaten

humblehacker:
This fixes the problem mentioned in Discussion #322


humblehacker:
Closing in favor of issue #180

PR #328, opened 2023-10-04
Fix visionOS build on latest beta SDK

### LINKED ISSUE #327: `inputAccessoryView` and `inputAssistantItem` missing in latest visionOS SDK
Trying to build on the latest visionOS SDK in Xcode 15.1 beta:

```
Cannot override 'inputAccessoryView' which has been marked unavailable
'inputAssistantItem' is unavailable in visionOS
```

This is weird because in UTM, we use `inputAccessoryView` without the override and it works fine. However, I don't know how to deal with `inputAssistantItem`.

osy:
`inputAccessoryView` and `inputAssistantItem` are not available.

Fixes #327


migueldeicaza:

PR #329, opened 2023-11-05
fix crash when apple pencil touch the screen

lampmanyao:


migueldeicaza:

PR #335, opened 2024-02-05
Made terminal variable public, so custom implementation can be used

Kotivskyi:
Fixes problem when you can subclass MacTerminalView and can subclass Terminal, but you can not initialize your "CustomMacTerminalView" with "CustomTerminal"


migueldeicaza:
Good call.  I had not considered that.

light end up deprecating getTerminal()

PR #339, opened 2024-03-16
Support VisionOS

crabtree-michael:
This add visionOS support.


migueldeicaza:
Odd, wonder why I have it working without this on vision

Issue #330, opened 2023-11-07
TerminalView does not respect frame height on macOS Sonoma


nkleemann:
As you can see from the screen recording, the `LocalProcessTerminalView` exceeds the frame height when it's displayed text exceeds the vertical height of the frame.

I'm targeting macOS Sonoma, using SwiftUI. I can also confirm that the **same code** worked an hour ago on macOS Ventura. 

Small GIF showing the bug: [Screen Recording](https://imgur.com/a/ptwo69v)

How I'm creating the View:

```swift
struct SuperColliderProcessView: NSViewRepresentable {
    
    @EnvironmentObject var superColliderProcessManager: SuperColliderProcessManager

    func makeCoordinator() -> Coordinator {
        Coordinator(self)
    }

    func makeNSView(context: Context) -> LocalProcessTerminalView {
        let terminalView = LocalProcessTerminalView(frame: CGRect(x: 0, y: 0, width: 300, height: 90))        
        return terminalView
    }

    func updateNSView(_ nsView: LocalProcessTerminalView, context: Context) {}

    class Coordinator: NSObject {
        var parent: SuperColliderProcessView
        init(_ parent: SuperColliderProcessView) {
            self.parent = parent
        }
    }
}

// From SwiftUI

SuperColliderProcessView()
    // even with these two, on or the other - the terminalview doesnt respect the set height
    .frame(width: 300, height: 90)
    .frame(maxHeight: 90)
```

Edit:
Another angle on this bug:

[Screen Recording](https://imgur.com/a/JdeFLHs)

Maybe it's linked to some Scroll Behaviour changes in Sonoma?

nkleemann:
Another angle on this bug:

[Screen Recording](https://imgur.com/a/JdeFLHs)

Maybe it's linked to some Scroll Behaviour changes in Sonoma?

nkleemann:
@migueldeicaza I found the commit that introduces this Bug in Sonoma, its the deadlock fix here: https://github.com/migueldeicaza/SwiftTerm/commit/e2c374952c77f5f43db7765f5e9f0cf47604c7b0 

Maybe a problem related to SwiftUI's drawing system, in relation to background threads?

Here is the self-contained example to recreate this bug. On macOS Sonoma, run this on any `SwiftTerm` version 1.2.2 and up:

```swift
import SwiftUI
import SwiftTerm

struct ContentView: View {
    
    @EnvironmentObject var replViewManager: REPLViewManager
[...]

PR #343, opened 2024-05-17
Fix for TerminalView growing out of bounds in Sonoma (#330)

nkleemann:
View clipping needs to be set explicitly in App/UIKit on Sonoma, see here: https://indiestack.com/2023/06/view-clipping-sonoma/

This PR just adds the flag to `TerminalView`'s `setup` function.


migueldeicaza:

Issue #286, opened 2023-03-23
Messed up character width for some fonts


lukepistrol:
For some fonts the character width is a bit messed up (i.e. too wide).

SwiftTerm:

<img width="502" alt="Screenshot 2023-03-24 at 00 30 59" src="https://user-images.githubusercontent.com/9460130/227387135-b0748b53-bc12-45dc-a290-55862aff7f69.png">

Apple's Terminal (iTerm looks the same):

<img width="347" alt="Screenshot 2023-03-24 at 00 32 05" src="https://user-images.githubusercontent.com/9460130/227387164-d320d631-c679-4038-9dfc-cc8b630a2e57.png">

The font used in this example is [FiraCode](https://fonts.google.com/specimen/Fira+Code) but this happens with a couple of other fonts (Source Code Pro for Powerline, Fira Mono for Powerline) too which look totally normal in other Terminal emulators.

migueldeicaza:
This is odd, how did you get this setup?

migueldeicaza:
Folks, I could appreciate a complete bug report, something I can reproduce on my end.

Otherwise, I spend hours tracking down third party issues.   I am using those fonts and do not have that problem on iOS.

scrapp08:
Adding onto this, nerd fonts do not seem to render correctly.

What I'm getting:
<img width="921" alt="image" src="https://github.com/migueldeicaza/SwiftTerm/assets/105889363/4ed24c81-5547-40e7-9973-dfc6987c0591">


What is expected:
<img width="1485" alt="image" src="https://github.com/migueldeicaza/SwiftTerm/assets/105889363/3f8c9b3f-7751-4e7b-92a2-73d570758f99">

jwells89:
I’m seeing this issue with a few fonts as well.
<img width="687" alt="FontBug" src="https://github.com/migueldeicaza/SwiftTerm/assets/5009493/3c83f2b3-b5d3-4580-9bb9-68a36bfc7bc9">

Screenshot taken in my own project, but it’s reproducible in the MacTerminal sample project by:
1. Installing the attached version of the free Cascadia Code font
2. In `viewDidLoad()` in `ViewController.swift`, inserting `terminal.font = NSFont(name: "Cascadia Code", size: 13)!`
3. Running the app

[...]

PR #347, opened 2024-08-07
Use W Hack For macOS Cell Width

thecoolwinter:
Uses the "W" hack for finding the cell width on macOS, similar to how it already works on iOS. Adds a commented version that comes with a performance hit as discussed in #286.


migueldeicaza:
Thank you for the PR!

PR #348, opened 2024-08-08
Fix Caret Drawing Artifacts

thecoolwinter:
Updates the caret drawing function to clip to it's bounds. Fixes a small visual bug where the unfocused box stuck around in the layer because it was never cleared due to it being outside the view's bounds. Fixes by first clearing the entire layer and then drawing.

Updates the macOS Caret background color when the color is set using the public variable. Fixes an issue where the caret color was not being used on macOS if it was changed after initialization.

**Screenshots**
Caret drawing is mostly the same, just that it's only drawing inside the bounds of the cell the cursor is in.

Caret drawing without patch, note leftover stroke artifacts from the unfocused box when focused:
Screenshot 2024-08-08 at 10 27 27 AM
Screenshot 2024-08-08 at 10 26 59 AM
Screenshot 2024-08-08 at 10 27 03 AM
Screenshot 2024-08-08 at 10 27 30 AM

Caret drawing with patch:
macOS from CodeEdit
Focused, block
Screenshot 2024-08-08 at 9 59 29 AM
Screenshot 2024-08-08 at 9 59 25 AM
Focused, underline
Screenshot 2024-08-08 at 9 58 43 AM
Screenshot 2024-08-08 at 9 58 40 AM
Focused, bar
Screenshot 2024-08-08 at 9 58 32 AM
Unfocused
Screenshot 2024-08-08 at 9 59 33 AM
Screenshot 2024-08-08 at 9 59 38 AM
iOS from example SSH app
IMG_8129
IMG_8133
IMG_8132
IMG_8131

### DIFF
--- a/Sources/SwiftTerm/Apple/CaretView.swift
+++ b/Sources/SwiftTerm/Apple/CaretView.swift
@@ -16,12 +16,14 @@ extension CaretView {
         guard let terminal else {
             return
         }
+        context.saveGState()
+        context.clip(to: [bounds])
         context.setFillColor(TTColor.clear.cgColor)
         context.fill ([bounds])
         
         if !hasFocus {
             context.setStrokeColor(bgColor)
-            context.setLineWidth(2)
+            context.setLineWidth(3)
             context.stroke(bounds)
             return
         }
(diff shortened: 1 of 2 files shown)


migueldeicaza:

PR #350, opened 2024-09-09
Changed executable to executableTarget

Vaidios:
Using `executable` is deprecated. This PR replaces that with newer API.

There was also this CaptureOutput target commented out, which I decided to also transfer to newer API. Is there a reason that it is commented out?


migueldeicaza:
CaptureOutput is really just a test for me, did not want to pollute thing for people.

Going to apply your patch manually, due to that secondary change.

Issue #212, opened 2022-01-18
No configuration option to set scrollback


ktprograms:
There is a `scrollback:` value in the `TerminalOptions` struct, but no public API has the ability to change it from the default of 500.

migueldeicaza:
Fixed

PR #356, opened 2024-11-26
Made options and setup methods public, so terminal options can be cus…

Kotivskyi:
Made options and setup methods public, so terminal options can be customized
scrollback now can be customized

Resolves this issue: [No configuration option to set scrollback](https://github.com/migueldeicaza/SwiftTerm/issues/212)


migueldeicaza:
Thank you!

PR #357, opened 2024-12-02
Improved performance with large scrollback

Kotivskyi:
Improved performance for terminal will large scrollback values (over 10000 lines), so memory will not be allocated for buffers that are not used

- Implemented optimisation for BufferLine, so data array will not be allocated until it is needed
- modified resize method to be more efficient if data are not allocated yet

Problem background:
During initialization and window resize Terminal will call resize method for each buffer line during which triggers memory reallocation of BufferLine.data. This operation becomes bottleneck with terminal scrollback > 10000.

Screenshot 2024-12-02 at 17 57 25

Implemented fix takes into account that most of the buffers will contains only blank charactes (until something useful will be printed into it) and can be lazy allocated using known fillCharacter and expectedDataSize. In this case resize method will be as simple as updating expectedDataSize value.

### DIFF
--- a/Sources/SwiftTerm/BufferLine.swift
+++ b/Sources/SwiftTerm/BufferLine.swift
@@ -25,26 +25,26 @@ public class BufferLine: CustomDebugStringConvertible {
     var renderMode: RenderLineMode = .single
     lazy var data: [CharData] = {
         isDataInitialised = true
-        return Array(repeating: fillData, count: dataColumns)
+        return Array(repeating: fillCharacter, count: expectedDataSize)
     }()
     
     var isDataInitialised = false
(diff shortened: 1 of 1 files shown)


Kotivskyi:
Note: Further optimization is possible since most of the lines in CircularList<BufferLine> will be empty ones and we don't really need to allocated those BufferLines before they contain anything useful.

migueldeicaza:
Thank you very much, this is great!

Another option is to only allocate the rows on demand, rather than allocating up-front and having the the scroll back be a limit to how much it grows, rather than a startup allocation size.

But I will merge your change until the day I fix that

PR #358, opened 2024-12-03
Fixed CI

Kotivskyi:
- bumped xcode version to 15.4, since 14.2 is not available on macos-latest (mac os 15) runner
- fixed iOS simulator build.: disabled arm64 architecture build, since SwiftSH does not supports it


Kotivskyi:
There is failing unit tests set SwiftTermTests that has dependency esctest which I assume is present on your local environment.
Do you think it will be fair to disable them when executing github CI?

migueldeicaza:
Mhm, esctest should be public, I think it would be nice to bring into the CI rather than removing it.

PR #362, opened 2025-01-16
re-setup scroller on size change

krzyzanowskim:
## problem
  NSScroller does not resize automatically (in height) after the initial frame setup.

## solution
  Re-setup scroller on resize. Unfortunately, simply updating NSScroller.frame.height does not work as expected.


migueldeicaza:
Thank you for the change Marcin!

PR #365, opened 2025-04-09
Fix issue where setting the caret color did nothing

georges:
Looks like internal variable was not being set.


migueldeicaza:
Thank you!

PR #367, opened 2025-04-20
Fix terminal clear on resize

georges:
On resize in iOS, terminal would clear which is a problem. Replaced hard reset with soft reset as suggested in https://github.com/migueldeicaza/SwiftTerm/discussions/366


migueldeicaza:
Thank you!

Issue #320, opened 2023-08-09
Voice dictation not working with terminal


karanmoon:
I try to speak command using voice dictation from keyboard it does not insert into terminal but whatever command I speak it inserts the blank space. in xcode console I'm getting below error. let me knock any way to overcome on this. I need to insert the spoken text into terminal.

Error : 
![Screenshot 2023-08-09 at 8 07 37 PM](https://github.com/migueldeicaza/SwiftTerm/assets/122771380/f2d39277-a29c-43c4-b8da-5dfd363fe955)

@migueldeicaza

karanmoon:
Can anybody have idea how to achieve this? @krzyzanowskim @migueldeicaza

krzyzanowskim:
I may have any idea how to poke it https://twitter.com/krzyzanowskim/status/1670483353185787906, can't promise I can find time to work on it now by myself tho

georges:
+1

georges:
I have an experimental branch with a refactor of iOSTextInput that seems to fix dictation. User gets hypothesis and correction sent to terminal until dictation is finalized. I will kick the tire on this some more before I open a pull request. 

https://github.com/islandmagic/SwiftTerm/tree/refactor-text-storage

PR #369, opened 2025-06-07
Refactor text input to handle dictation

georges:
* Refactored text storage to use String and fixed several text input method implementations.
* Fixed [dictation bug](https://github.com/migueldeicaza/SwiftTerm/issues/320): interim dictation text is now displayed as speech is recognized, and replaced with the final result when dictation ends.
* Fixed backspace behavior: properly deletes large chunks after a certain repeat threshold.
* Improved double-tap selection to be more inclusive of common characters found in file names.


migueldeicaza:
Whoa, this is amazing!   I am going to review it.

migueldeicaza:
Thank you for this incredible contribution!

PR #380, opened 2025-07-04
Add support for setting child process working directory in LocalProcess

krzyzanowskim:
Allows specifying a custom currentDirectory when launching a process. Modifies PseudoTerminalHelpers.fork to accept an optional working directory parameter and calls chdir() before execve().

LocalProcess.startProcess() now propagates this new parameter.


migueldeicaza:
Great change!

PR #381, opened 2025-07-15
Consider Buffer Y Offset In Mouse Hit

thecoolwinter:
Fixes a bug where selecting text when the terminal was scrolled would select the text as if the terminal was not scrolled.

Before: 

https://github.com/user-attachments/assets/7446776b-fbe2-4176-a756-09c1fadf2dad

With changes:

https://github.com/user-attachments/assets/fe38189b-0a19-4c0e-a46d-b22666090832


migueldeicaza:

PR #382, opened 2025-07-15
Update Scroller When Exists - Fix Layout Exception

thecoolwinter:
Fixes a layout exception due to the removal of the scroller during constraint layout. This happened when the view was embedded in a SwiftUI view and SwiftUI attempted to update its frame, which caused the scroller to be removed from the view hierarchy which is illegal.

To fix I just removed the removal of the old scroller and updated its frame if it's not `nil`. This avoids a view tree update and still updates the scroller's position and size.


migueldeicaza:
Thank you for your fix!

PR #384, opened 2025-07-17
Fix Selected Text

thecoolwinter:
The selection service's start/end positions can be reversed (usually when dragging backwards in the text). This orders the start/end positions before calling `terminal.getText`.


migueldeicaza:
Thank you for the patch!

PR #387, opened 2025-08-12
fix wrong backgroun color of wide unicode char

hanyang-tony:
The cell width is fixed, and to display wide characters, we render both the character and an empty space character to avoid them mashed together.

However, the empty space character's background color is inverted, resulting in this cursed look.
Capture d’écran 2025-08-12 à 14 16 25

If we are sure the character is wide, add an empty space with same background color. The "wide" character set is copied from iTerm2.
Capture d’écran 2025-08-12 à 14 19 47


migueldeicaza:
Thank you for the contribution!

I think that one part is a no-brainer to add, the rendering itself.

The second part poses two problems: (a) I am not sure that we need this, shouldn't this just be a call to wcswidth() and if the width is 2, it is a long-character?   (b) iTerm2 is licensed under the GPL or LGPL and bringing code from there would make the combined work for SwiftTerm be GPL/LGPL, but SwiftTerm is MIT licensed.

migueldeicaza:
I provided an alternate implementation using a version of wcwidth based on SpectreKit/TerminalGui code, and also had to rework the loop there, as otherwise, pasting wide attributes would render them at the beginning of the line, rather than inline.

PR #388, opened 2025-08-12
fix wrong selection range in scroll mode

hanyang-tony:
The display offset is incorrectly added to selection range, if the buffer line is larger than actual display height (display offset large than 0), the row selected is the actual row plus display offset.

Stop adding the offset, so the row selected is the row mouse pointed at.


migueldeicaza:

Issue #161, opened 2021-01-05
Mac Terminal View shows scroller


aaronvegh:
I've been using SwiftTerm for my Mac app, injecting a TerminalView. Recently I noticed that there's a white square on my view's background:

<img width="550" alt="Screen Shot 2021-01-05 at 4 09 31 PM" src="https://user-images.githubusercontent.com/34525/103699201-99312b80-4f70-11eb-8cda-1baab519d857.png">

I realized it was a scrollbar. Although this is apparently a vertical scrollbar, it's rendering a horizontal one that moves left/right as you scroll up/down, and while the thing has a correct frame, the rendered version only uses the intrinsic height and the width of the bar. Weird!

[I found where you're creating the scroller](https://github.com/migueldeicaza/SwiftTerm/blob/master/Sources/SwiftTerm/Mac/MacTerminalView.swift#L253), and experimentation revealed that _not adding the scroller to the view_ obviously removed the problem, while also keeping the ability to scroll intact.

<img width="637" alt="Screen Shot 2021-01-05 at 4 10 05 PM" src="https://user-images.githubusercontent.com/34525/103699461-f62ce180-4f70-11eb-8980-2c2b784411eb.png">

I wonder what you think of this approach? I wonder if the use of the legacy scrollbar (compare to Terminal.app's use of the overlay scrollbar) is appropriate? If I had a choice of whether to show a funny-looking square scroller, or no scroller that still scrolled, I'd choose the latter, but I wonder if I'm screwing something up and this is just me. I can never overlook that possibility.

Thanks so much for this wonderful package.

A.

migueldeicaza:
This was part of syncing the features that were being added to XtermSharp by the Visual Studio team to the Swift version:

https://github.com/migueldeicaza/XtermSharp/commit/9581543faf558989451ddf0ce03666153ae636fd

When I ported the code, I vaguely remember struggling with that, and seeing it, and fixing it.   Or at least, I thought I had fixed it.    Does this just happen when embedding, or also when using the off the shelf terminal?

aaronvegh:
I just ran the standalone Mac Terminal, and the scroller works "fine", in that it's the legacy scroller and shows up okay:

<img width="912" alt="Screen Shot 2021-01-06 at 10 23 56 AM" src="https://user-images.githubusercontent.com/34525/103785551-553e3500-5009-11eb-8707-b0db74abe1c4.png">

Indeed, the .overlay scroller doesn't look right either!

[...]

PR #389, opened 2025-08-27
Temporary fix scroller issue https://github.com/migueldeicaza/SwiftTe…

hmduc1603:
I know this issue was opened years ago and recent it happened in my latest SwiftUI Mac Project, so I decided to make a temporary fix for those whon need.

https://github.com/migueldeicaza/SwiftTerm/issues/161

Technically I set fixed height of NSScroller by initializing TerminalView via a variable (scrollerHeight), so when setupScroller() get called, it used that height instead of bounds.height.

`public init(scrollerHeight: CGFloat) {}`


sgehrman:
still no fix? a working scrollbar seems like the most obvious thing to support.

migueldeicaza:
Proper solution implemented.

PR #396, opened 2025-09-14
Fix a build issue on LocalProcess with Environment

ipodishima:
Build was failing on LocalProcess L495 with `Cannot convert value of type '[String : String]' to expected argument type '[Environment.Key : String]'`


migueldeicaza:
Thank you for your contribution!   I am guessing the API changed slightly since I first adopted it.

ipodishima:
Yes and because the package doesn’t block a version, i wasn’t able to use main :)

PR #403, opened 2025-11-14
Fix(iOS): ensure soft keyboard backspace works after paste

bernie-du:
The software keyboard's backspace key failed to send a keypress if the internal `textInputStorage` buffer was empty. This problem always occurred following paste operations, as the paste operation sends text directly to the terminal without populating this buffer.

This commit fixes the issue by modifying the `deleteBackward` method to send a backspace character even when `textInputStorage` is empty, aligning with user expectations.

Additionally, the method has been updated to consistently respect the `backspaceSendsControlH` setting, ensuring its behavior is identical to the hardware keyboard implementation.


migueldeicaza:

I had just fixed it too, and then came to check the open PRs, but your fix is better than mine, so I undid my change and merged yours.

bernie-du:
Wow, thank you, I'm honored to make a small contribution to this library.

Issue #407, opened 2025-12-06
Unable to enter characters using Chinese or Japanese keyboards on iOS


holzschu:
**Describe the bug**

When entering Chinese or Japanese text on iOS, using the on-screen keyboard, glyphs selected on the keyboard do not appear on screen.

**To Reproduce**

- switch to a Chinese/Japanese keyboard
- type letters, e.g. f-u-j-i 
- ideograms appear on the toolbar. Click on one of them.
- the ideogram does not appear on screen.

**Expected behavior**

The ideogram you clicked should appear on the screen.

**Screenshots**

Observed behaviour: 

https://github.com/user-attachments/assets/464bf5ac-7a1c-4af9-9d3c-9dc0dcff5f9a

Expected behaviour: 

https://github.com/user-attachments/assets/3f89664a-6cd0-42e4-8fb9-504944e539ac

**Desktop (please complete the following information):**
 - OS: OSX
 - Version SwiftTerm release 1.5.1 and also using the latest commit ( 3630da9441c7207392bf7dba02eb1452085f4858 )

**Smartphone (please complete the following information):**
 - Device: iPhone SE and iPad Pro
 - OS: iOS 26.2 and iOS 16.7.2

**Additional context**

A fix for this issue would require to have #406 and #402 also fixed (rendering of wide characters).

PR #409, opened 2025-12-07
Allow keyboard input on iOS for languages with multi-char input (Japanese, Chinese, Korean, Hindi)

holzschu:
This is potential fix for #407 (entering Chinese, Japanese, Korean ... characters from the keyboard on iOS), with a minimal set of changes to the code.

It works in two steps: 
- in `iOS/iOSTerminalView.swift`, do not treat the keyboard event if the keyboard language is one of "ja", "zh", "hi" or "ko", and let the system handle it.
- in `iOS/iOSTextInput.swift`, when the user has clicked on a character or group of characters, `unmarkText()` is called and we get the string to insert from `_markedRange`.

The languages themselves work in different ways: Chinese and Japanese will bring a menu with a set of choices, depending on what the user has typed, while Korean with send an "deleteBackward()" event when two characters should be combined (type "d" followed by "n" on a US external keyboard to see what I mean, or use an on-screen Korean iOS keyboard and type "ㅇ" followed by "ㅜ". The result should be: "우").

There is another language with a specific input method, "Tiếng Việt Télex", and that one is still not working with these changes. The way it works (in other apps, like Pages) is that you type "Tieng Viet Telex" and iOS adds the diacritics depending on the context. But *how* that works is still a mystery for me (there are no keyboard events visible).

### DIFF
--- a/Sources/SwiftTerm/iOS/iOSTerminalView.swift
+++ b/Sources/SwiftTerm/iOS/iOSTerminalView.swift
@@ -1330,8 +1330,14 @@ open class TerminalView: UIScrollView, UITextInputTraits, UIKeyInput, UIScrollVi
                 } else if key.modifierFlags.contains (.alternate) && optionAsMetaKey {
                     data = .text("\u{1b}\(key.charactersIgnoringModifiers)")
                 } else if !key.modifierFlags.contains (.command){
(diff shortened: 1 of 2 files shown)


migueldeicaza:
Thank you for looking at this!

Merged

Issue #406, opened 2025-12-06
Emojis and wide characters have a white rectangle masking half of them on iOS


holzschu:
**Describe the bug**

When entering a wide character on iOS, the right half of the character is obscured by a white rectangle (see picture)

**To Reproduce**

Open the emoji keyboard, enter any emoji character. The behaviour is worse with compose emojis (see #341 ) who are separated into multiple components, each with their own white rectangle.

**Expected behavior**

I would expect the emojis to appear on screen, without the white rectangle masking half of them.

**Screenshots**

The problem (running SwiftTerm on iOS):

<img width="375" height="667" alt="Image" src="https://github.com/user-attachments/assets/a828d391-cf8b-4c63-b17a-c5e50e49f0fb" />

The expected behaviour:

<img width="375" height="667" alt="Image" src="https://github.com/user-attachments/assets/87407abc-7754-4109-a233-11edb832f93d" />

**Desktop (please complete the following information):**
 - OS: OSX
 - Version: SwiftTerm release 1.5.1 and also using the latest commit (3630da9441c7207392bf7dba02eb1452085f4858)

**Smartphone (please complete the following information):**
 - Device: iPhone SE and iPad Pro
 - OS: iOS 26.2 and iOS 16.7.2
 - Browser N/A

**Additional context**

migueldeicaza:
Ouch, I think I regressed this recently.

holzschu:
I've tracked it (so far) to this line: https://github.com/migueldeicaza/SwiftTerm/blob/48beac28c956d8f114bbb08b87f255cc8a786823/Sources/SwiftTerm/Buffer.swift#L1140
If I comment out the line, the left half of wide characters is printed (but the right half is missing); if I leave it as is, we get the white rectangle. Earlier, `empty` was defined to:
```Swift
var empty = CharData.Null
```

I'm still investigating.

migueldeicaza:
Quick workaround, can you try to set 'nativeBackgroundColor' to UIColor.clear, rather than setting it directly?

migueldeicaza:
Ok, this should now be properly fixed in 607674b46265b198644a028e59caed25294b9019

holzschu:
Thank you for all those changes. It's much better now (and the skin color modifier works, which is important). I have spotted some emojis that only render half their width (at least ☯️ and ☦️, I don't see the logic yet).

My understanding is that the only way to know for sure the width of a character... is to compute the width of the character:
```Swift
[...]

Issue #341, opened 2024-04-20
Working with compose emojis such as "🖐🏾" or "👩‍👩‍👦‍👦"


holzschu:
**Describe the bug**
When I send a emoji made of several characters with compose, like "🖐🏾" or "👩‍👩‍👦‍👦", to terminalView using: `terminalView?.feed(text: string)`, it appears on the view as the concatenation of the characters, like this: "👨👩👦👧"

**To Reproduce**
Type "🖐🏾" or "👩‍👩‍👦‍👦" using the emoji keyboard, or call `terminalView?.feed(text: "🖐🏾")`

**Expected behavior**
I would expect the composed emoji to appear instead of the concatenation. 

**Additional context**
It may be that I am doing something wrong; maybe I should not call `terminalView.feed()` to insert text like composed emojis, but I could not find which other method I should call.

PR #410, opened 2025-12-09
Rendering emojis and wide characters on iOS

holzschu:
This is more a Request For Comment than a fully formed Pull Request. 

It solves my two issues with emojis and wide characters (#341 and #406), with minimal changes to the code, but not in the most elegant way. Those lines must have been there for a reason, but the only thing that works for me is to deactivate them.

#341 : compose emojis. Deactivate the line ` if firstScalar.properties.canonicalCombiningClass != .notReordered {` in `Sources/SwiftTerm/Terminal.swift` for iOS. On the iOS emoji keyboard, you can have a combined emoji whose first character is `notReordered` and must still be combined (🖐🏾 = combine + 🖐️ + skin color modifier, for example).

#406 : white squares on wide characters. The only solution that works on iOS is to deactivate the line `context.fill(rect.applying(transform))` in `Apple/AppleTerminalView.swift`, but this might have some side effects.

My goal is to help others who encounter similar issues with a quick workaround, until you have a better solution in the code.

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -724,7 +724,8 @@ extension TerminalView {
                     let rect = CGRect (origin: origin, size: size)
                     #if os(macOS)
                     rect.applying(transform).fill(using: .destinationOver)
-                    #else
+                    #elseif (false)
+                    // iOS: do not clear the background so characters are well rendered:
                     context.fill(rect.applying(transform))
                     #endif
                     context.restoreGState()
--- a/Sources/SwiftTerm/Terminal.swift
+++ b/Sources/SwiftTerm/Terminal.swift
@@ -1200,7 +1200,12 @@ open class Terminal {
 
             if let firstScalar = ch.unicodeScalars.first {
                 // If this is a Unicode combining character
-                if firstScalar.properties.canonicalCombiningClass != .notReordered {
+                var alwaysCombine = false
+                #if os(iOS)
+                // iOS specificity: always go through this loop for combine Unicode
+                alwaysCombine = true
+                #endif
+                if alwaysCombine || firstScalar.properties.canonicalCombiningClass != .notReordered {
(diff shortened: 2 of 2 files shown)


migueldeicaza:
Thanks for this detective work!   It turns out that the rules for the combining are a lot more elaborate than I knew!

migueldeicaza:
Mhm, the problem with disabling the call to the context.fill() call is that it prevents the background colors from rendering, see for example 'mc'.

Investigating.

migueldeicaza:
Ok, I think that the issue is the computation of the next element to render:

```
                var positions = runGlyphs.enumerated().map { (i: Int, glyph: CGGlyph) -> CGPoint in
                    CGPoint(x: lineOrigin.x + (cellDimension.width * CGFloat(col + i)), y: lineOrigin.y + yOffset)
                }
....
                    let transform = CGAffineTransform (translationX: positions[0].x, y: 0)

```

The latter is not taking into consideration the width - and I thin that the previous hack was wrong

migueldeicaza:
Come to think of it - the previous hack only worked because it forced an extra "hidden" space there, but the positioning of these glyphs was also wrong, it is not just the background that is off.

migueldeicaza:
Ok, this renders now correcty and undoes the last change.

https://gist.github.com/migueldeicaza/ee25a96a20ea8a838db5bf2e759379b5

But I think it is too costly, I think a better approach is to keep track of the cumulative usage per run, but the problem is that it is not easy to map runs to the column/usage at that point,

migueldeicaza:
Update: I do not think there is any sane way of making the CTRun-based system match the actual columns, so this is a deeper problem than a quick rendering fix - it fundamentally has a disagreement on the position - one is what the Terminal itself builds, and one is what CTRun extracts.

I think the right fix is to have buildAttributedString to return segments of the same width, and then apply rendering based on those groups, rather than attempt to have those match each other.

holzschu:
Thanks for all the changes. I'm gonna close this PR.

PR #412, opened 2025-12-11
Adding more 2-columns emojis (mostly religious symbols)

holzschu:
If I understood correctly the code in `Sources/SwiftTerm/Utilities.swift`, this adds a few extra emojis to the two-column list, based on: https://en.wikipedia.org/wiki/Religious_and_political_symbols_in_Unicode

Some of them were already present (e.g. ⛪️) but not all (⛩️, ☯️...)

⛩️ (`0x26e9`) is actually in a special class: it can either be one-column (⛩) or two-columns (⛩️) depending on the unicode "variation selector" that follows: `0x26e9 0xfe0e` = ⛩, `0x26e9 0xfe0f` = ⛩️. Globally, any unicode character followed by `0xfe0e` will be single column, any unicode character followed by `0xfe0f` will be double-column :
https://en.wikipedia.org/wiki/Variation_Selectors_(Unicode_block)

I'm not sure how to deal with those special cases with the existing code in `columnWidth()`, but if you can give me some directions, I'll try.

### DIFF
--- a/Sources/SwiftTerm/Utilities.swift
+++ b/Sources/SwiftTerm/Utilities.swift
@@ -133,6 +133,12 @@ struct UnicodeUtil {
         LH (lo: 0x25fd, hi: 0x25fe),
         // ☔, ☕
         LH (lo: 0x2614, hi: 0x2615),
+        // religious and political symbols:
+        LH (lo: 0x2626, hi: 0x2626),
+        LH (lo: 0x262a, hi: 0x262a),
+        LH (lo: 0x262c, hi: 0x262c),
+        LH (lo: 0x262e, hi: 0x262f),
+        LH (lo: 0x2638, hi: 0x2638),
         // zodiac signs
         LH (lo: 0x2648, hi: 0x2653),
         // ♿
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Oh, this is very interesting!

I added special handling for 0xfe0f to mean two columns - I think it is a bit of a stretch from the Wikipedia page, but like everything related to this, I think it is acceptable.

migueldeicaza:
Ok, landed my change that should cope with the two sizes.

migueldeicaza:
Ok, with my new change of merging, I had to undo one small bit of your patch.

That will make it so that the 2/1 sizes are respected based on the modifier doing bw/color (from the wiki page), and default to bw/single column otherwise.

migueldeicaza:
The one part I am unclear on from your patch is: how many of those need to be undone now that I have support for 0xfe0f

holzschu:
Your change makes sense, I was actually planning to write a new PR to specifically undo ⛩️ now that there is support for 0xfe0f.

About your other question (what needs to be undone now that you have support for 0xfe0f), that is more challenging.
I found some reference here: https://en.wikipedia.org/wiki/Dingbats_(Unicode_block)

- first, I was wrong: a character can be followed by 0xfe0e and still be two columns (but anything followed by 0xfe0f is definitely two columns).
- second, there's a handful of characters listed there with an actual one-column glyph: ✂︎, ✌︎...
- but third: ⛩ is not on that list, so it's not a complete definite list...

holzschu:
I still have an issue with 0x2e69 / ⛩️: with the latest commit (43e92a785b010a960d12db7cebaa21f86e5b9401), it displays as 2 columns, but the cursor only advances by one column, so the next character is hidden (letters) or becomes a big rectangular block (emojis).

migueldeicaza:
Oh that part I know exactly what is going on - while we now compute the right size, when updating I was not changing the next column value.

Fix incoming.

PR #416, opened 2025-12-18
Fix: macOS background color rendering using incorrect color

yanjunhui:
On macOS, NSRect.fill(using:) uses NSColor (set via NSColor.set() or setFill()), not the CGContext's fill color that was set via context.setFillColor().

This caused background colors to be rendered incorrectly - for example, when running `ls --color` in a terminal, files that should have colored foreground text would instead appear with that color as background.

The fix adds backgroundColor.setFill() before calling rect.fill() to ensure the correct NSColor is used for background rendering.


migueldeicaza:

PR #417, opened 2025-12-23
MacLocalTerminalView: Add currentDirectory parameter to startProcess()

claimhawk:
The underlying `LocalProcess.startProcess()` already supports a `currentDirectory` parameter for specifying the working directory when launching a process, but `MacLocalTerminalView.startProcess()` did not expose this parameter.

This required users to access the internal `process` property directly:

```swift
// Required workaround:
terminalView.process.startProcess(executable: shell, args: ["--login"], currentDirectory: homeDir)
```

With this change, the API is consistent:

```swift
// Clean API:
terminalView.startProcess(executable: shell, args: ["--login"], currentDirectory: homeDir)
```

**Changes:**
- Added `currentDirectory: String? = nil` parameter to `MacLocalTerminalView.startProcess()`
- Updated documentation to describe the new parameter
- Passes the parameter through to the underlying `process.startProcess()` call

**Use case:** When building a terminal app that should start the shell in a specific directory (e.g., user's home directory rather than the app's working directory), this parameter allows clean configuration without accessing internal properties.


migueldeicaza:

PR #420, opened 2026-01-05
Add programmatic rendering for Unicode block elements

adammeghji:
## Summary
- Implement custom drawing for block characters (U+2580-U+259F) to eliminate anti-aliasing artifacts at cell boundaries
- Add `BlockDrawing.swift` with programmatic rendering for all block elements (full/half blocks, eighth blocks, shades, quadrant characters)
- Modify `AppleTerminalView` to detect block characters and render them with `CGContext` fills instead of `CTFontDrawGlyphs`

This is the same approach used by Kitty, WezTerm, and other modern terminals to ensure crisp block character rendering.

## Problem
When rendering block characters using font glyphs via CoreText, anti-aliasing causes visible gaps/banding at cell boundaries. This is particularly noticeable in TUI applications displaying waveforms, progress bars, or any UI using block elements.

## Solution
Detect block characters before calling `CTFontDrawGlyphs` and render them programmatically using `CGContext.fill()` with anti-aliasing disabled. Non-block characters continue to use normal font rendering.

## Test plan
- [x] Builds successfully with `swift build`
- [x] Tested in macOS app with TUI displaying waveforms using block characters
- [x] Verified no banding/gaps at cell boundaries

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -799,16 +799,51 @@ extension TerminalView {
 
                     nativeForegroundColor.set()
 
+                    var fgColor = nativeForegroundColor.cgColor
                     if runAttributes.keys.contains(.foregroundColor) {
                         let color = runAttributes[.foregroundColor] as! TTColor
                         let cgColor = color.cgColor
+                        fgColor = cgColor
                         if let colorSpace = cgColor.colorSpace {
                             context.setFillColorSpace(colorSpace)
                         }
                         context.setFillColor(cgColor)
                     }
-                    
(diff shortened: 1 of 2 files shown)


migueldeicaza:
Interesting idea.

Can you share a screenshot before and after, just so I know what I am merging?

adammeghji:
> Interesting idea.
> 
> Can you share a screenshot before and after, just so I know what I am merging?

Absolutely, let me work on exactly this so you'll see the difference in rendering when the banding shows vs when it doesn't.

migueldeicaza:
Ok, I see what this does.

I took a slightly different approach, and instead of hijacking the text during the CoreTExt rendering, I render it in a pass like the kitty graphics.   Please take it for a spin.

migueldeicaza:
New: 
image

Old:

image

PR #422, opened 2026-01-06
Fix CJK character cursor drift caused by font advance width mismatch

fuwasegu:
## Summary

Fixes cursor position drifting when typing CJK (Japanese, Chinese, Korean) double-width characters.

## Problem

When typing CJK characters, the cursor gradually drifts away from the text. The gap between the cursor and the actual text grows larger with each character typed.

## Steps to reproduce

1. Open a terminal using SwiftTerm
2. Type Japanese text: echo あいうえお
3. Observe the cursor position drifting further from the text with each character

## Root cause

In AppleTerminalView.swift, glyph positions were calculated using CoreText's font-based advance width:

```swift
positions[i] = CGPoint(
    x: ctPosition.x + xOffset,
    y: ...)
```

CoreText's glyph advance width may not exactly match cellDimension.width × 2 for double-width characters, causing cumulative offset as more characters are rendered.

## Solution

Calculate each glyph's X position based on the terminal's column grid instead of CoreText's relative positioning:

```swift
let glyphColumn = startColumn + (i * segment.columnWidth)
positions[i] = CGPoint(
    x: lineOrigin.x + CGFloat(glyphColumn) * cellDimension.width,
    y: ...)
```

This ensures all characters align correctly with the terminal's cell grid regardless of font metrics.

## Testing

- Added unit tests for CJK character positioning:
  - testCJKCharacterPositioning - Japanese hiragana
  - testCJKMixedWithAscii - Mixed ASCII and Japanese
  - testChineseCharacterPositioning - Chinese characters
- All existing tests pass


migueldeicaza:
This was an incredible patch - I am not sure why my comments got deleted before, but I love that this also solves the Emoji drift problem and works great with Arabic!

Thank you so much for finding the spot to do this.

Issue #309, opened 2023-06-13
Mac: auto-scroll up and down while dragging selection


migueldeicaza:
This might have been a regression, it does not auto-scroll the buffer when the mouse is dragged out of the window.

migueldeicaza:
Fixed in f971cbf6cf19c8ff94ceb8bc310b758e6036e626

PR #430, opened 2026-01-09
Fix auto-scroll and SIGINT/Ctrl+C on macOS

botsdown:
This PR contains two fixes for macOS:

  ## 1. Fix auto-scroll timer for drag selection

  When drag-selecting text past the terminal edges, the view should auto-scroll. This fix addresses multiple issues:

  1. Added timer management:
     - autoScrollTimer property to hold the timer reference
     - startAutoScrollTimer() to create timer with .common run loop mode
     - stopAutoScrollTimer() to invalidate and clean up
     - Timer start/stop calls in mouseDragged() and cleanup in mouseUp()

  2. Fixed scroll direction in scrollingTimerElapsed():
     - Was calling scrollUp() in both branches
     - Now correctly calls scrollDown() when autoScrollDelta > 0

  3. Added selection extension during scroll:
     - Calls selection.dragExtend() after scrolling to include newly visible rows

  4. Fixed scroll-down detection in mouseDragged():
     - calculateMouseHit() clamps row to valid range, so scroll-down condition was never triggered
     - Now uses raw pixel coordinates from mouse event to detect when mouse is outside view bounds

  Uses .common run loop mode to ensure timer fires during mouse tracking.

  ---

  ## 2. Fix SIGINT/Ctrl+C by forcing forkpty on macOS

  ### Problem

  On modern macOS with swift-subprocess available, `LocalProcess.startProcess()` uses the `startProcessWithSubprocess()` path. This breaks Ctrl+C (SIGINT) because:

  1. The Subprocess path uses `POSIX_SPAWN_SETSID` which creates a new session **without a controlling terminal**
  2. Without a controlling terminal, the kernel's line discipline doesn't convert ETX (0x03) to SIGINT
  3. `tcgetpgrp(childfd)` returns -1 (ENOTTY) because there's no controlling terminal
  4. `shellPid` is never set in the Subprocess path (remains 0)

  The result: pressing Ctrl+C shows `^C` but doesn't interrupt the running process.

  ### Solution

  Force the `forkpty` path on macOS. The `forkpty` function internally calls `login_tty()` which:
  - Calls `setsid()` to create a new session
  - Calls `ioctl(TIOCSCTTY)` to set the PTY slave as the controlling terminal
  - Properly sets up stdin/stdout/stderr

  This is a minimal change - just a `#if os(macOS)` conditional to prefer `forkpty` on macOS while preserving the Subprocess path for other platforms.

  ### Testing

  1. Run a terminal with SwiftTerm on macOS
  2. Execute `sleep 100`
  3. Press Ctrl+C
  4. **Before fix**: `^C` appears but sleep continues
  5. **After fix**: Process terminates immediately, `echo $?` shows 130 (128 + signal 2)

  ### References

  - [The TTY demystified](https://www.linusakesson.net/programming/tty/) - Explains controlling terminal requirements
  - [PTY man page](https://man7.org/linux/man-pages/man7/pty.7.html) - Documents signal generation requirements

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,0 +1,32 @@
+{
+  "pins" : [
+    {
+      "identity" : "swift-argument-parser",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/apple/swift-argument-parser",
+      "state" : {
+        "revision" : "c5d11a805e765f52ba34ec7284bd4fcd6ba68615",
+        "version" : "1.7.0"
+      }
+    },
+    {
+      "identity" : "swift-subprocess",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/swiftlang/swift-subprocess",
+      "state" : {
+        "branch" : "main",
+        "revision" : "426790f3f24afa60b418450da0afaa20a8b3bdd4"
+      }
+    },
+    {
+      "identity" : "swift-system",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/apple/swift-system",
+      "state" : {
+        "revision" : "395a77f0aa927f0ff73941d7ac35f2b46d47c9db",
+        "version" : "1.6.3"
+      }
+    }
+  ],
+  "version" : 2
+}
--- a/Sources/SwiftTerm/LocalProcess.swift
+++ b/Sources/SwiftTerm/LocalProcess.swift
@@ -233,7 +233,11 @@ public class LocalProcess {
             return
         }
         
-        #if canImport(Subprocess)
+        // Force forkpty on macOS for proper controlling terminal (SIGINT) support
+        // The Subprocess path doesn't set up a controlling terminal, breaking Ctrl+C
+        #if os(macOS)
+        startProcessWithForkpty(executable: executable, args: args, environment: environment, execName: execName, currentDirectory: currentDirectory)
+        #elseif canImport(Subprocess)
         startProcessWithSubprocess(executable: executable, args: args, environment: environment, execName: execName, currentDirectory: currentDirectory)
         #else
         startProcessWithForkpty(executable: executable, args: args, environment: environment, execName: execName, currentDirectory: currentDirectory)
--- a/Sources/SwiftTerm/Mac/MacTerminalView.swift
+++ b/Sources/SwiftTerm/Mac/MacTerminalView.swift
@@ -927,17 +927,50 @@ open class TerminalView: NSView, NSTextInputClient, NSUserInterfaceValidations, 
     }
     
     private var autoScrollDelta = 0
+    private var autoScrollTimer: Timer?
+
     // Callback from when the mouseDown autoscrolling timer goes off
     private func scrollingTimerElapsed (source: Timer)
     {
         if autoScrollDelta == 0 {
             return
         }
+
+        let displayBuffer = terminal.displayBuffer
+
         if autoScrollDelta < 0 {
+            // Dragging above top edge - scroll up (show earlier content)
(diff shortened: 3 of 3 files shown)


migueldeicaza:
How does Control-C work, I am using it just fine, can you show me what seems to be the problem?

migueldeicaza:
The auto-scroll seems sensible, I think I can add that.

migueldeicaza:
Btw, I tried the reproduction steps you have, and for me, I get exactly that without your change:

image

botsdown:
Interesting. Let me revisit that one and make sure I had the latest version
downloaded before testing again. Thanks!

botsdown:
After more research, I think the difference might be which code path
SwiftTerm takes at compile time.






The current logic is:




#if canImport(Subprocess)



    startProcessWithSubprocess(...)



#else



    startProcessWithForkpty(...)



#endif







swift-subprocess was added to the Swift toolchain relatively recently. If
you're on an older Xcode/Swift version where canImport(Subprocess)
evaluates to false, you'd get the forkpty path automatically. Ctrl+C works
fine there because forkpty establishes a proper controlling terminal.







In my particular environment, Xcode 16.2 / Swift 6, canImport(Subprocess)
is true, so we take the Subprocess path.

The issue I am seeing is that the Subprocess implementation:



1. Creates a PTY pair with openpty()
2. Sets POSIX_SPAWN_SETSID to create a new session
3. But never calls ioctl(TIOCSCTTY) to establish the PTY as the controlling
terminal

Without a controlling terminal, the kernel's line discipline doesn't know
where to send SIGINT when it receives ETX (0x03). The character gets
written to the PTY, but no signal is generated.


Could you check which path you're hitting? Adding this would confirm:

#if canImport(Subprocess)
print("Using Subprocess path")
#else
print("Using forkpty path")
#endif

If you are on the Subprocess path and Ctrl+C still works for you, There may
be something else I'm missing or something unique about my environment.

migueldeicaza:
I am on the aubprocess path

botsdown:
Well, not sure what the difference is then. Let me dig a little deeper
tomorrow. It must be something on my end. Thanks, Miguel!

migueldeicaza:
I bet I know what it is.

Did you remove the Sandbox entitlement?   I bet the sandbox is failing.

botsdown:
I don't think so... but I will check on that today!

migueldeicaza:
Any updates?

mcglothi:
Independently reproduced this exact bug in a packaged macOS app (portside) and confirmed your diagnosis: \`autoScrollDelta\`/\`scrollingTimerElapsed\` in \`MacTerminalView.swift\` compute the right thing but nothing ever schedules a \`Timer\` to drive it, so drag-selection past the top/bottom edge never scrolls.

This looks like the same bug reported in #533 and #309 (both closed without a fix actually landing — the dead code is still present in current \`main\`), so there's real, repeated demand for this. Your fix here (timer lifecycle in mouseDragged/mouseUp, the scrollUp/scrollDown direction bug, extending the selection during the scroll, and the pixel-coordinate fix for detecting below-view drags since calculateMouseHit clamps) matches everything I found by hand.

Worked around it downstream for now with an NSEvent monitor since we can't wait indefinitely, but would much rather delete that workaround once this or an equivalent fix is merged. Hope this helps get it unstuck — happy to help test if a rebase resolves the conflicts.

botsdown:
Thanks for the update. I had to move away from Swift for my project, but
appreciate the info.

migueldeicaza:
Folks, the timer code was added in commit f971cbf I think that this is issue is now fixed.

Please reopen if you still see this problem, I am unable to reproduce now.

Issue #256, opened 2022-09-28
Crash on reverseIndex


migueldeicaza:
I caught the issue in the wild, the crash is:

```
    frame #3: 0x000000019475f134 libswiftCore.dylib`Swift._assertionFailure(_: Swift.StaticString, _: Swift.String, file: Swift.StaticString, line: Swift.UInt, flags: Swift.UInt32) -> Swift.Never + 652
    frame #4: 0x0000000102b7f84c LaTerminal`CircularList.shiftElements(start=47, count=3, offset=1, self=0x0000000280bf5680) at CircularList.swift:169:9
  * frame #5: 0x0000000102beebe4 LaTerminal`Terminal.reverseIndex(self=0x00000001043e4ed0) at Terminal.swift:4937:26
    frame #6: 0x0000000102bbeb98 LaTerminal`closure #84 in Terminal.configureParser(collect=0 values, flag=77, self=0x00000001043e4ed0) at Terminal.swift:868:76
```

The issue is that `shiftElements` is being called with start=47, but the buffer itself only has 42 lines.   The 47 is a combination of buffer.y = 21 and buffer.base = 26.     And also notice that buffer.y = buffer.scrollTop which is why this path is triggered at all, originMode=false, usingMargins() = false

I was running top, and messing up the display with debug output from a background process, and I had refreshed it a few times, and I had just attempted to go to the background with control-z

The data was:
  [0] = "\r"
  [1] = "\u{1b}"
  [2] = "["
  [3] = "3"
  [4] = "4"
  [5] = "d"
  [6] = "\u{1b}"
  [7] = "["
  [8] = "L"
  [9] = "\u{1b}"
  [10] = "["
  [11] = "2"
  [12] = "2"
  [13] = ";"
  [14] = "2"
  [15] = "5"
  [16] = "r"
  [17] = "\u{1b}"
  [18] = "["
  [19] = "2"
  [20] = "2"
  [21] = ";"
  [22] = "1"
  [23] = "H"
  [24] = "\u{1b}"
  [25] = "M"
}

Cursor Vertical Absolute Position 34
Insert Line 
Set Top and Bottom Margin to 22;25 (sets buffer.scrollBottom = 24, buffer.scrollTop=21)
Set Cursor Position 22;1
Reverse Index

Most likely, what we have here are two different buffers in action.

migueldeicaza:
Fixed in recent contribution

PR #431, opened 2026-01-10
Fix crash on reverseIndex and cmdRestoreCursor (issue #256)

PhilipLudington:
## Summary

- Fix crash in `reverseIndex()` when `yBase` exceeds `lines.count` (issue #256)
- Fix crash in `cmdRestoreCursor()` when `savedX`/`savedY` are invalid after resize
- Add defensive bounds checks in `scroll()` for the same class of bug

## Root Cause

`Buffer.clear()` was not resetting `yBase` to 0. When switching between normal and alternate buffers, stale `yBase` values caused `buffer.y + buffer.yBase` to exceed `buffer.lines.count`, triggering out-of-bounds access.

Similarly, `cmdRestoreCursor()` restored saved cursor positions without bounds checking, causing `abort()` in Debug builds when saved values became invalid after resize/scroll operations.

## Changes

- **Buffer.swift**: Add `yBase = 0` to `clear()` to fix the root cause
- **Terminal.swift**: 
  - Add defensive bounds checks in `reverseIndex()` and `scroll()`
  - Clamp `savedX`/`savedY` to valid ranges in `cmdRestoreCursor()`
- **BufferTests.swift**: Add 12 unit tests covering the bugs and edge cases

## Test plan

- [x] All 12 new buffer tests pass
- [x] Full test suite passes (62 tests)
- [x] Tests verify the fix catches the bug (fail without fix, pass with fix)

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/Sources/SwiftTerm/Terminal.swift
+++ b/Sources/SwiftTerm/Terminal.swift
@@ -2428,8 +2428,10 @@ open class Terminal {
 
     func cmdRestoreCursor (_ pars: [Int], _ collect: cstring)
     {
-        buffer.x = buffer.savedX
-        buffer.y = buffer.savedY
+        // Clamp savedX and savedY to valid ranges to prevent abort() in Debug builds.
+        // Saved values can become invalid after resize/scroll operations.
+        buffer.x = min(max(0, buffer.savedX), cols - 1)
+        buffer.y = min(max(0, buffer.savedY), rows - 1)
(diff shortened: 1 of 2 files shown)


migueldeicaza:
Wow, thank you for fixing this very long-standing bug!

PhilipLudington:
You are welcome!  Thanks for building an awesome tool!

PR #432, opened 2026-01-10
fix: Add periodic Kitty image garbage collection during scroll

kaantradelab:
Problem:
When using Kitty graphics protocol, orphan images accumulate in `kittyGraphicsState.imagesById` because `cleanupUnusedKittyImages()` is only called on explicit delete commands.

During long-running terminal sessions with heavy output (e.g., CI/CD logs, AI agent sessions), this causes memory to grow unbounded. Observed: 11GB+ memory usage, terminal freeze, inability to scroll.

Solution:
- Add counter-based garbage collection in `scroll()`
- Track lines scrolled since last cleanup
- Trigger `cleanupUnusedKittyImages()` every 500 lines when buffer is trimmed
- Change function visibility from `private` to `internal`

Results:
- Memory usage reduced from 11GB+ to ~400MB (96%+ improvement)
- Terminal remains responsive during long sessions

### DIFF
--- a/Sources/SwiftTerm/KittyGraphics.swift
+++ b/Sources/SwiftTerm/KittyGraphics.swift
@@ -1703,7 +1703,9 @@ extension Terminal {
         return right >= screenLeft && left <= screenRight && bottom >= screenTop && top <= screenBottom
     }
 
-    private func cleanupUnusedKittyImages() {
+    /// Removes images from `imagesById` that are no longer referenced by any buffer line.
+    /// Called periodically during scroll to prevent memory accumulation.
+    func cleanupUnusedKittyImages() {
         var used = Set<UInt32>()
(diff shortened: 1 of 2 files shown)


migueldeicaza:
Oops, thank you so very much for tracking this!

Let me see if I can do something a little more aggressive, rather than doing a cycle scan every N scrolled lines, if that does not pan out, I will merge your PR.

migueldeicaza:
Btw, the reason I do not want to do a collection after the number of lines is both that it would add another layer of uncertainty and jumps while I am profiling the performance and because it would probably break the kitty protocol.

Since reference counting does not work (as this is part of the design), what I am going to do is add an eviction of data after certain size.

migueldeicaza:
Ok, this is what I am doing instead:

https://github.com/migueldeicaza/SwiftTerm/pull/433

PR #434, opened 2026-01-10
Fix scroll() to respect left/right margins (DECSLRM)

leftspin:
## Summary

When margin mode is active with narrowed left/right margins (DECSLRM), the `scroll()` function was scrolling entire lines instead of only the columns within the margin region. This caused display corruption when applications like tmux use DECSLRM for side-by-side panes - scrolling in one pane would affect content in adjacent panes.

The fix adds a new code path that detects when:
1. Margin mode is enabled (DECLRMM)
2. Margins are narrower than full width
3. The scroll should only affect the margin region

In this case, we do in-place column-by-column scrolling within the margin bounds, similar to how `cmdInsertLines`/`cmdDeleteLines` already handle this scenario.

### Additional Fixes

- **cmdIndex (ESC D)**: Now checks if cursor is within left/right margins before triggering scroll, matching the behavior of `cmdLineFeedBasic()`
- **reverseIndex (ESC M)**: Same margin check added, plus margin-aware column-by-column scrolling for narrow margins

## Known Limitation

Line-level metadata (`isWrapped`, `images`, `renderMode`) cannot be partially scrolled since they are per-`BufferLine` properties spanning all columns. The fix resets these on affected lines. A comment in the code documents that ideally `isWrapped` would be tracked per-column-range for proper triple-click selection within individual panes, but this would require architectural changes to the data model.

## Test Plan

- [x] Tested with tmux side-by-side panes - scrolling in one pane no longer corrupts adjacent pane
- [x] Verified normal scrollback behavior preserved when margins are full width
- [x] Built and ran in visionOS app (Floatty)
- [x] Tested intermittent "text pushed down" issue in adjacent pane - fixed by cmdIndex/reverseIndex changes


migueldeicaza:
This is great and makes an existing failure in the test suite pass (test_DECRQSS_DECSLRM), but it comes with an additional change that is not necessary (the Package.resolved), would you mind removing that file from the PR?

leftspin:
oops. sloppy. will do

migueldeicaza:
Thank you!

PR #448, opened 2026-01-16
Fix NO-BREAK SPACE (U+00A0) column width calculation

kuzmeech:
## Summary

NO-BREAK SPACE (U+00A0) was incorrectly treated as non-printable (width -1) because the condition `irune <= 0xA0` included it in the C1 control character range. This caused cursor positioning issues when applications use NBSP in their output.

## Changes

Changed the condition from `irune <= 0xA0` to `irune < 0xA0` in `columnWidth()` so that U+00A0 falls through to the default return value of 1, matching the expected behavior of a space character.

## Use Case

Claude Code (Anthropic's terminal-based AI coding assistant) uses NO-BREAK SPACE after its prompt character (`❯`). With the incorrect width calculation, typing the first character after the prompt would cause cursor misalignment and visual glitches (duplicate/ghost characters).

## Testing

- [x] All 155 existing tests pass
- [x] Manual testing with Claude Code - first character no longer duplicates
- [x] Verified U+00A0 now returns width 1 instead of -1

### DIFF
--- a/Tests/SwiftTermTests/UnicodeTests.swift
+++ b/Tests/SwiftTermTests/UnicodeTests.swift
@@ -274,5 +274,30 @@ final class SwiftTermUnicode {
         #expect(line == "\(sequence)X")
     }
 
+    @Test func testNoBreakSpaceWidth() {
+        let h = HeadlessTerminal (queue: SwiftTermTests.queue) { exitCode in }
+        let t = h.terminal!
+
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Let me first validate with ucs-detect

migueldeicaza:
Good news, that passed.

PR #452, opened 2026-01-29
Fix cursor ghosting in TUI applications

eyelock:
## Problem

TUI applications like Claude Code were displaying ghost cursors - white cursor artifacts remaining visible at previous positions when the cursor moved.

## Root Cause

The `updateCursorPosition()` method in AppleTerminalView was not removing CaretView when `terminal.cursorHidden == true`. TUI apps use DECTCEM escape sequences (`ESC[?25l`/`ESC[?25h`) to hide/show the cursor during rendering, and the missing removal logic caused CaretView instances to accumulate.

## Solution

Added logic to remove CaretView from the view hierarchy when the cursor is hidden:

```swift
} else if terminal.cursorHidden == true && caretView.superview == self {
    caretView.removeFromSuperview()
    return
}
```

## Changes

- **File**: `Sources/SwiftTerm/Apple/AppleTerminalView.swift`
- **Function**: `updateCursorPosition()`  
- **Lines Added**: 4 (lines 1206-1209)

## Testing

See `CURSOR_GHOSTING_BUG.md` for detailed manual testing procedure.

**To test manually:**
1. Build SwiftTerm with this fix
2. Run a TUI app like Claude Code in TermQ
3. Verify no ghost cursors remain when typing
4. Compare with behavior before fix (checkout HEAD~1)

## Regression Risk

**Low** - Change is localized to cursor visibility handling in AppleTerminalView only.

## Fixes

Resolves cursor ghosting issue when running TUI applications (Claude Code, htop, vim, etc.) that use cursor hide/show sequences.

### DIFF
--- a/CURSOR_GHOSTING_BUG.md
+++ b/CURSOR_GHOSTING_BUG.md
@@ -1,0 +1,121 @@
+# Cursor Ghosting Bug Fix
+
+## Problem
+
+TUI (Text User Interface) applications like Claude Code, htop, and other ncurses-based apps would display "ghost cursors" - white cursor artifacts that remained visible at previous positions when the cursor moved.
+
+## Root Cause
+
+The `updateCursorPosition()` method in `Sources/SwiftTerm/Apple/AppleTerminalView.swift` was not removing the CaretView from the view hierarchy when the cursor was hidden.
+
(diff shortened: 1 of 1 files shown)


eyelock:
Closed - created against wrong repository. This PR should target eyelock/SwiftTerm fork for review before upstream submission.

Issue #405, opened 2025-12-01
Window Resize on Coding Agents (eg. Claude Code or Gemini CLI)


roromedia:
**Describe the bug**
Whenever you resize the window to a smaller size than the initial loading size, the terminal doesn't correctly reflow. This leads to double-lining and endless scrolling while typing. As soon as you enlarge the window over the initial size, this stops, but the prompt box isn't resized to the max-width of the window.

**To Reproduce**
Just resize window as mentioned above.

**Expected behavior**
It should reflow the contents and always max-width the prompt block. Terminals like Rio or Alacritty are doing this perfectly.

**Desktop (please complete the following information):**
 - OS: MacOS Tahoe 26.1

**Additional context**
I tried to fork SwiftTerm and implement various reflow mechanisms but couldn't make it work. Seems there are  more substantial or underlying changes needed.

migueldeicaza:
Sounds like a task for Claude!

tdrozdowski:
Add Junie to the list (Jetbrains) - not sure what's going on but I cannot get it to render properly in SwiftTerm at all.  Claude works well for me (mostly) - so found this surprising.

tdrozdowski:
I took a stab at this with Claude Code.  No luck.  I wound up rolling my project back to v1.5.0 and it behaves perfectly.  So not entirely sure what changed between all of these (need to look more closely at all the commits) - but something did.

btw, 1.6.0 doesn't build cleanly as you have a dependency on a pre-release item that fails my build.  I had to pin it to the git hash to get it to build.  But then I ran into this issue so...back to no cut-n-paste, but the AI CLIs at least function (except for Junie - no idea what witchcraft their TUI is using to not work properly).

roromedia:
I also tried with Claude Code before opening this issue - also no luck.
Alacritty which is a Rust-based terminal, as well as the new JetBrains terminals work well or at least better.

migueldeicaza:
The 1.5.0 is a good data point

What’s the issue in 1.6.0, I should fix it

tdrozdowski:
Here's the error I get: Dependencies could not be resolved because package 'swiftterm' is required using a stable-version but 'swiftterm' depends on an unstable-version package 'swift-subprocess' and root depends on 'swiftterm' 1.6.0.

On macOS 26.3 Beta + Xcode 26.2

tdrozdowski:
FWIW - Claude's analysis on the diff between 1.5.0 and 1.6.0:

[...]

PR #455, opened 2026-02-01
Push window-size changes to the child process group and handle the swift-subprocess path

cchen140:
This commit contains the following changes:
- added a single resize path that updates the PTY size and sends SIGWINCH, in Sources/SwiftTerm/LocalProcess.swift.
- hooked terminal view resize to that new path in Sources/SwiftTerm/Mac/MacLocalTerminalView.swift.
- captured the subprocess PID when using swift-subprocess, so SIGWINCH can be delivered in that path too.

As a result of these changes, $COLUMNS/$LINES can be updated correctly when terminal view is being resized.

### DIFF
--- a/Sources/SwiftTerm/LocalProcess.swift
+++ b/Sources/SwiftTerm/LocalProcess.swift
@@ -297,15 +297,21 @@ public class LocalProcess {
                         posix_spawnattr_setflags(&spawnAttr, flags | Int16(POSIX_SPAWN_SETSID))
                         
                     }
-                    let result = try await Subprocess.run(
-                        .name(executable),
+                    let configuration = Subprocess.Configuration(
+                        executable: .name(executable),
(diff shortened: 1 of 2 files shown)


cchen140:
This should solve the issue raised in #405.

migueldeicaza:
Interesting find.   This patch generally looks good, but I do not think that we need to send SIGWINCH, the kernel should send that in response to us setting the value.

Were you able to reproduce the original issue?   I would to validate that this is fixing it.

migueldeicaza:
I tested with and without your patch and both `stty` and bash's COLUMN/ROWS work the same before and after, so I am not sure what this is fixing.

cchen140:
I was able to reproduce the original issue in my own macOS app integration (which is why this commit). In my case I wrapped the terminal in a view in SwiftUI, so possibly the issue was derived from integrating with SwiftUI.

migueldeicaza:
Can you try the new version?   I pushed a change that should update the values correctly.

PR #460, opened 2026-02-03
Make HeadlessTerminal.process public

zxss702:
The `process` property in `HeadlessTerminal` was internal, making it inaccessible from external modules that import SwiftTerm. This prevented users from calling `startProcess()` on the `LocalProcess` instance.

Since `HeadlessTerminal` is designed to provide a headless terminal emulator that can be scripted and screen-scraped, users need access to the underlying `LocalProcess` to start a child process after initialization.

This change aligns with the existing pattern in the package's own `CaptureOutput` target, which also accesses the `process` property directly (but works because it's in the same package).


migueldeicaza:
Thanks!

PR #466, opened 2026-02-12
Fix keyboard toggle flicker on iOS

alvarolb:
## Problem

When toggling between the system keyboard and the custom accessory keyboard on iOS using `iOSAccessoryView`, there was a visible flicker. The old code called `resignFirstResponder()` + `becomeFirstResponder()` which triggered a full keyboard dismiss/show animation cycle.

## Fix

Replace `resignFirstResponder()`/`becomeFirstResponder()` with `reloadInputViews()`, which swaps the input view in place without dismissing the keyboard. This eliminates the flicker entirely.

The change is minimal: just the toggle method in `iOSAccessoryView.swift`.


migueldeicaza:
Thanks, this looks good.

PR #467, opened 2026-02-12
Clamp savedY to valid bounds on terminal resize

buzzboxmedia:
## Summary

- `Buffer.resize()` clamps `savedX` to `newCols - 1` but does not clamp `savedY`, allowing it to exceed the new row count after resize
- `Terminal.resizeBuffers()` can produce a negative `savedY` when recomputing the offset after resize (`normalBuffer.y + dy` where `dy` was captured before resize)
- Both cases cause `abort()` in the `Buffer.y` setter on DEBUG builds when `cmdRestoreCursor` assigns `buffer.y = buffer.savedY`

## Fix

- Clamp `savedY` to `[0, newRows - 1]` in `Buffer.resize()`, matching the existing `savedX` clamp
- Clamp `savedY` to `>= 0` in `Terminal.resizeBuffers()` after recomputing the offset

## Test plan

- [x] Verify terminal resize no longer crashes when a saved cursor position exceeds new row count
- [ ] Run existing SwiftTerm test suite

### DIFF
--- a/Sources/SwiftTerm/Buffer.swift
+++ b/Sources/SwiftTerm/Buffer.swift
@@ -486,6 +486,7 @@ public final class Buffer {
             }
 
             savedX = min (savedX, newCols - 1)
+            savedY = max (min (savedY, newRows - 1), 0)
 
             scrollTop = 0
         }
--- a/Sources/SwiftTerm/Terminal.swift
+++ b/Sources/SwiftTerm/Terminal.swift
@@ -760,7 +760,7 @@ open class Terminal {
         // correct the savedY cursor to follow changes to y
         let dy = normalBuffer.savedY - normalBuffer.y
         normalBuffer.resize (newCols: newColumns, newRows: newRows)
-        normalBuffer.savedY = normalBuffer.y + dy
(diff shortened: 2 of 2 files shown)


migueldeicaza:
Mhm, I think that this has shown that there might be a bit of a misuses with yDisp, I think this might need a deeper fix.

migueldeicaza:
I think that this called into question a much bigger problem, which is that currently the code mixes savedY as either a viewport row (0..<rows) and sometimes an absolute buffer row (yBase-based) - and with these differences what you found was a step in the right direction, but not complete.

I compared this to xterm.js which was the original inspiration, and they do not have that discrepancy - either it got fixed there, or I introduced the bug.

This patch:

https://gist.github.com/migueldeicaza/fb7470de4b4c9501c7eb660bc56a3ede

Now treats savedY as absolute (yBase-baesd).  Also, it now converts absolute savedY back to viewport row (savedY - yBase) and then clamps via restrictCursor().

While auditing I found another difference, softReset now also resets the saved cursor position

Would you mind taking this for a spin?

migueldeicaza:
I created a branch `absolute-savedY`, I need to test that on iOS as well, but I am tied up for the next couple of days.

PR #468, opened 2026-02-12
Fix iOS scroll locked to bottom during incoming data

alvarolb:
## Problem

On iOS, `TerminalView` (a `UIScrollView` subclass) always forces `contentOffset` to the bottom in `updateScroller()`. This means the user cannot scroll up to read previous output while the terminal is actively receiving data — any new content immediately snaps the view back to the bottom.

## Approach

This follows the same pattern already used on macOS (where `updateScroller()` only updates the `NSScroller` widget without forcing content position), adapted for `UIScrollView`:

1. **Derive contentOffset from yDisp** — `updateScroller()` now uses `displayBuffer.yDisp` (already managed by `Terminal.scroll(isWrapped:)`) instead of always forcing to the bottom. When `userScrolling` is true, `yDisp` stays stable so the view won't snap.

2. **Track scroll via panGestureRecognizer** — Uses `addTarget` on the existing pan gesture instead of setting `delegate = self`, preserving delegate semantics for consumers of `TerminalView`.

3. **Synchronize yDisp from contentOffset** — After the user finishes scrolling (including deceleration), `yDisp` is updated to match the visual scroll position, keeping buffer state in sync.

4. **Detect deceleration end via CADisplayLink** — Uses the existing display link to check when deceleration ends, avoiding `contentOffset` `didSet` overrides that conflict with UIKit bounce animations.

5. **Reset on buffer switch** — `bufferActivated()` clears `userScrolling` so alternate screen transitions always snap to bottom.

6. **Reset on keyboard input** — `commitTextInput()` clears `userScrolling` and scrolls to bottom, since typing implies the user wants to see current output.

7. **Conditional ensureCaretIsVisible** — Only scrolls to bottom if the caret is actually off-screen, matching macOS behavior.

## Changes

- `Sources/SwiftTerm/iOS/iOSTerminalView.swift`: Guard `contentOffset` in `updateScroller()`, add pan gesture tracking, sync `yDisp` from scroll position, reset on buffer switch and keyboard input, conditional `ensureCaretIsVisible`.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,0 +1,95 @@
+{
+  "pins" : [
+    {
+      "identity" : "hdrhistogram-swift",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/HdrHistogram/hdrhistogram-swift.git",
+      "state" : {
+        "revision" : "de0b9b8a27956b9bfc9b4dce7d1c38ad7c579f19",
+        "version" : "0.1.4"
+      }
+    },
+    {
+      "identity" : "package-benchmark",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/ordo-one/package-benchmark",
+      "state" : {
+        "revision" : "a1075611342d4b164bf6e977edb02a3db4434afd",
+        "version" : "1.29.11"
+      }
+    },
+    {
+      "identity" : "package-jemalloc",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/ordo-one/package-jemalloc.git",
+      "state" : {
+        "revision" : "e8a5db026963f5bfeac842d9d3f2cc8cde323b49",
+        "version" : "1.0.0"
+      }
+    },
+    {
+      "identity" : "swift-argument-parser",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/apple/swift-argument-parser",
+      "state" : {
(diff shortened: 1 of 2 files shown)


migueldeicaza:
I like the idea, but there are a few challenges with this patch

1. The patch toggles terminal.userScrolling but never synchronizes yDisp from the actual UIScrollView position.   I think this will get the data out of sync.
2. I think that the state reset needs to cover other scenarios, like buffer switching 
3. Hijacking the delegate is a change in the semantics, I am not sure how to solve this.
4. I wonder what happens if you are just a little off, it seems like this requires one full line to be scrolled before the "freeze" effect takes place.

migueldeicaza:
Oh, and I think that keyboard input needs to also auto-reset this.

migueldeicaza:
I got a GitHub notification for this, but when I click on it, it says nothing is there  - would you mind posting your message again?

MartinStrambach:
Hey guys, any updates on this issue? This is especially annoying when I use Claude Code and want to scroll to previous messages, but new data keep coming in.

migueldeicaza:
I think a few folks have forks with band-aids, but they suffer from the above problems, so I would love to get this feature, but without introducing errors.

migueldeicaza:
Ok, can you folks try this branch; prevent-scroll-while-dragging

MartinStrambach:
@migueldeicaza works for me 🥳

migueldeicaza:
Ok, I merged that branch.   Let me know if there is something else we need merging.

PR #469, opened 2026-02-12
Fix NSScroller layout when TerminalView uses Auto Layout constraints

### LINKED ISSUE #161: Mac Terminal View shows scroller

colemaco:
Fixes #161                                                                                                            
                                                                                                                        
  ## Problem                                                                                                            
                                                                                                                        
  When `TerminalView` is embedded using Auto Layout constraints (e.g., via SwiftUI's `NSViewRepresentable` with         
  `translatesAutoresizingMaskIntoConstraints = false`), the NSScroller appears mispositioned - stuck in the bottom-right
   corner, scrolling horizontally instead of vertically.                                                                
                                                                                                                        
  ## Root Cause                                                                                                         
                                                                                                                        
  The scroller was positioned using `autoresizingMask` (frame-based layout), which conflicts with constraint-based      
  layout. The frame calculations in `getScrollerFrame()` relied on `bounds.maxX` which may not be accurate before the   
  layout pass completes.                                                                                                
                                                                                                                        
  ## Solution                                                                                                           
                                                                                                                        
  Use Auto Layout constraints to position the scroller:                                                                 
                                                                                                                        
  - Scroller is pinned to trailing/top/bottom anchors of the parent view                                                
  - Width uses the system-provided `NSScroller.scrollerWidth()` value                                                   
  - Works correctly whether parent uses frame-based or constraint-based layout                                          
                                                                                                                        
  ## Changes                                                                                                            
                                                                                                                        
  - `setupScroller()` now uses Auto Layout constraints instead of autoresizing masks                                    
  - `updateScrollerFrame()` is now a no-op (constraints handle positioning)                                             
  - Added `scrollerWidth` computed property for reliable width calculation                                              
  - `getScrollerFrame()` preserved for backwards compatibility                                                          
                                                                                                                        
  ## Testing                                                                                                            
                                                                                                                        
  Tested with SwiftUI/NSViewRepresentable embedding. Scroller now appears correctly on the right edge and scrolls       
  vertically as expected. 

Thanks for maintaining SwiftTerm - it's been great to use in my project.

### DIFF
--- a/Sources/SwiftTerm/Mac/MacTerminalView.swift
+++ b/Sources/SwiftTerm/Mac/MacTerminalView.swift
@@ -386,6 +386,7 @@ open class TerminalView: NSView, NSTextInputClient, NSUserInterfaceValidations, 
     }
 
     let scrollerStyle: NSScroller.Style = .legacy
+
     func getScrollerFrame() -> CGRect {
         let scrollerWidth = NSScroller.scrollerWidth(for: .regular, scrollerStyle: scrollerStyle)
         return NSRect(x: bounds.maxX - scrollerWidth, y: 0, width: scrollerWidth, height: bounds.height)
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Oh Autolayout that might explain why I am not able to reproduce that bug, probably folks are hosting the code and reporting bugs against that.   Thanks for your patch, I will review it shortly.

migueldeicaza:
Ok, reviewed, this is great, cheers!

migueldeicaza:
Thank you for tracking this down and figuring out that the issue was the different style of hosting.

colemaco:
Glad I could contribute - appreciate all the work you put into this.

PR #471, opened 2026-02-20
Fix: preserve text selection when mouse reporting is disabled

ThinkerJack:
## Problem

When `allowMouseReporting` is set to `false`, users expect to manually select text in the terminal. However, both `feedPrepare()` and `linefeed()` unconditionally clear the active selection on every data feed.

This causes **selection flickering and loss** in high-throughput scenarios — for example, when running streaming CLI tools (like AI coding assistants) that continuously write to the terminal. The user selects text, but it's immediately cleared by the next incoming data chunk.

## Fix

Guard both selection-clearing calls behind `allowMouseReporting`:

- `AppleTerminalView.feedPrepare()`: only set `selection.active = false` when mouse reporting is enabled
- `MacTerminalView.linefeed()`: only call `selection.selectNone()` when mouse reporting is enabled

This preserves the existing behavior when mouse reporting is on (selection is managed by the application), while allowing manual selections to persist when mouse reporting is off.

## Context

We discovered this while embedding SwiftTerm in [GroAsk](https://groask.com) (a macOS menu bar AI launcher) to run Claude Code sessions. The terminal receives high-frequency output, and users need to select and copy text while output is streaming.

Really appreciate your work on SwiftTerm — it's an incredibly well-architected terminal engine. The clean separation between the core emulator and platform-specific views made it straightforward to integrate. Thank you for open-sourcing it! 🙏

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -1570,7 +1570,12 @@ extension TerminalView {
     func feedPrepare()
     {
         search.invalidate()
-        selection.active = false
+        // Only clear selection when mouse reporting is enabled.
+        // When mouse reporting is disabled, the user is manually selecting
+        // text and expects it to persist across incoming data.
+        if allowMouseReporting {
+            selection.active = false
+        }
         startDisplayUpdates()
     }
     
--- a/Sources/SwiftTerm/Mac/MacTerminalView.swift
+++ b/Sources/SwiftTerm/Mac/MacTerminalView.swift
@@ -462,7 +462,12 @@ open class TerminalView: NSView, NSTextInputClient, NSUserInterfaceValidations, 
     }
     
     open func linefeed(source: Terminal) {
-        selection.selectNone()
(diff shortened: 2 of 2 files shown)


migueldeicaza:
Are there other terminals that have this behavior?   This does not sound like a good default to have.

ThinkerJack:
Yes! The terminals I've tested: Terminal.app, iTerm2, and Alacritty all preserve selection while output is streaming. xterm too — they even had a [patch](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=291787) fixing accidental clearing back in 2005.

The only one that doesn't is Kitty, which clears selection on any content change for performance reasons.

If you'd prefer this as an opt-in option rather than changing the default, I'm happy to add a flag for it.

migueldeicaza:
I just tested this, you are correct.   I wonder where I got the impression that this was not the case.

Ok, let me review the change.

migueldeicaza:
Ok, this is missing iOS support, and I had to test locally, so I will commit my local version momentarily.

migueldeicaza:
Thank you for your contribution!   I landed my locally tested version.

PR #473, opened 2026-02-20
Fix iOS compile error: wrap cmdBackTab in .bytes()

MehdiG44:
v1.11.0 introduced `SendData` enum for iOS key handling, but the shift-tab case assigns `EscapeSequences.cmdBackTab` (`[UInt8]`) directly to a `SendData?` variable without wrapping it in `.bytes()`.

This causes a compile error on iOS:
```
cannot assign value of type '[UInt8]' to type 'TerminalView.SendData'
```

One-line fix: wrap in `.bytes()` like all other key assignments in the same method.


migueldeicaza:
Thank you!

MehdiG44:
Thanks for the quick merge and release\! Small issue: the `V1.11.1` tag has an uppercase `V` — SPM can't resolve it as a semver version. Previous tags (`v1.11.0`, `v1.10.1`, etc.) use lowercase `v`. Could you re-tag with lowercase?

migueldeicaza:
Oops!   Fixed

PR #475, opened 2026-02-22
Make cellDimension, CellDimension, and BufferLine.isWrapped public

MehdiG44:
## Summary

Makes three properties public so embedding apps can implement tap-to-open-URL and other position-aware features:

- `TerminalView.cellDimension` (`iOSTerminalView.swift`) — needed to convert tap coordinates to terminal rows/columns
- `CellDimension` typealias (`AppleTerminalView.swift`) — required because `cellDimension`'s type must also be public
- `BufferLine.isWrapped` (`BufferLine.swift`) — needed to distinguish soft-wrapped lines from hard newlines

## Motivation

Terminal apps commonly need to detect URLs under a user's tap. This requires:
1. Converting a UIKit tap point to a terminal row (needs `cellDimension`)
2. Joining soft-wrapped lines to reconstruct full URLs (needs `isWrapped`)

These are read-only access patterns — no behavior changes, no API surface increase beyond visibility.

## Changes

Three `internal` → `public` visibility changes. No logic changes, no new API.

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -57,7 +57,7 @@ struct ViewLineInfo {
 }
 
 extension TerminalView {
-    typealias CellDimension = CGSize
+    public typealias CellDimension = CGSize
     
     func resetCaches ()
     {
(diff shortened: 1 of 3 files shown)


migueldeicaza:
I have been working on proper URL handling, both auto-detected and properly annotated URLs.

Would you mind checking my branch `dingus` and tell me if that is sufficient for your needs or what we could do to adjust it?

The reason is that once these things become public, they limit what I can do internally without breaking people.

migueldeicaza:
The URL handling landed, and the public version of the isWrapped is there, do we still need cell?

PR #476, opened 2026-02-22
Fix block element boundary check excluding U+2580 and U+259F

evertjr:
The block element range check uses strict inequality (`>` and `<`), which excludes U+2580 (UPPER HALF BLOCK) and U+259F from the custom block element renderer. These two characters fall through to font glyph rendering while all their neighbors (U+2581–U+259E) use the custom renderer.

This causes visible rendering artifacts — the font glyphs for U+2580 and U+259F have slightly different metrics than the custom-rendered blocks, so when they sit next to each other (which is very common in TUI applications like OpenCode, Claude Code, etc.), you get misaligned edges.

The box drawing path already uses inclusive bounds (`>=` and `<=`), so this just aligns the block element check to match.

image

**Before (U+2580 rendered by font, neighbors by custom renderer):**

Characters like ▀ show visual "horns" where the font glyph doesn't align with adjacent custom-rendered blocks.

**After (all block elements use custom renderer consistently):**

image

Seamless rendering across all block element characters.

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -183,7 +183,11 @@ extension TerminalView {
         let fontAttributes = [NSAttributedString.Key.font: fontSet.normal]
         let cellWidth = "W".size(withAttributes: fontAttributes).width
         #endif
-        return CellDimension(width: max (1, cellWidth), height: max (min (cellHeight, 8192), 1))
+        // Snap to pixel grid to avoid sub-pixel seams between adjacent cells
+        let scale = backingScaleFactor()
+        let snappedWidth = ceil(cellWidth * scale) / scale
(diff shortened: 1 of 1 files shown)


evertjr:
I also noticed that at certain column positions, adjacent block elements still had a faint seam , like a 1px misalignment visible on the "n" leg in OpenCode's logo, for example.                              

The root cause: computeFontDimensions returns fractional cell sizes (e.g. 7.8pt width for SF Mono at size 13). On a 2x Retina display that's 15.6 device pixels per cell, so some column boundaries fall between pixels. When two adjacent cells have different colors, the sub-pixel boundary creates a visible seam.                                                                                         

The second commit rounds cell dimensions up to the nearest device pixel using backingScaleFactor(), so every cell edge lands on an exact pixel. 

  Before (fractional cell width, seam on the "n" leg):

image

  After (pixel-snapped cell width):

image

migueldeicaza:
Wow great catch!

PR #478, opened 2026-02-26
Fix SGR 2 (dim/faint) rendering on light backgrounds

evertjr:
## Problem

`dimmedColor()` reduces each RGB component by 50 %, effectively blending toward black. On dark backgrounds this looks correct, but on light backgrounds it makes dim text darker.

A secondary issue: an alpha-based workaround (`withAlphaComponent(0.5)`) produces the right apparent color but leaves visible seams between adjacent box-drawing characters (e.g. `─────`) because each cell composites independently.
image

This problem is visible on Claude Code in light theme:

SwiftTerm:
image

macOS Terminal:
image

## Fix

Replace the no-arg `dimmedColor()` with `dimmedColor(towards:)` that takes the cell's background color and produces a fully-opaque 50 % blend between foreground and background. This:

- Reduces contrast correctly regardless of whether the theme is light or dark
- Eliminates tiling artifacts on box-drawing character runs
- Applies to both macOS (`NSColor`) and iOS (`UIColor`)

The call site in `AppleTerminalView` already computes the background color for each cell, so passing it to the new method is straightforward.

Here is how it looks after the fix:
image

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -397,14 +397,15 @@ extension TerminalView {
         }
         
         var fgColor = mapColor (color: fg, isFg: true, isBold: isBold, useBrightColors: useBrightColors)
-        // Apply dim/faint attribute (SGR 2) - reduce color intensity
-        if flags.contains(.dim) {
-            fgColor = fgColor.dimmedColor()
+        let bgColor = mapColor (color: bg, isFg: false, isBold: false)
+        // Apply dim/faint attribute (SGR 2)
+        if flags.contains (.dim) {
+            fgColor = fgColor.dimmedColor (towards: bgColor)
         }
         var nsattr: [NSAttributedString.Key:Any] = [
             .font: tf,
             .foregroundColor: fgColor,
-            .backgroundColor: mapColor(color: bg, isFg: false, isBold: false)
+            .backgroundColor: bgColor
         ]
         if flags.contains (.underline) {
             let underlineColor = attribute.underlineColor.map {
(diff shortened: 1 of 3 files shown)


evertjr:
I'll be testing for a while to see if this introduces any problems during use, when I'm confident I'll mark as ready for review.

migueldeicaza:
Perhaps we should just use alpha transparency in that case, rather than going to black.   That should give us the same effect, this is what Ghostty does.

evertjr:
> Perhaps we should just use alpha transparency in that case, rather than going to black. That should give us the same effect, this is what Ghostty does.

I updated the approach based on your feedback and dimmedColor() now uses alpha transparency, same as Ghostty.                                                                                       

One thing I ran into is that visible seams between adjacent box-drawing characters (e.g. ─────) like I mentioned in the PR. To handle that, I added opaqueComposite(over:) which pre-composites the color against nativeBackgroundColor before drawing in drawBoxDrawings and drawBlockElements. Regular text still uses natural alpha blending.

image

Tested on both light and dark themes, looks correct on both. Ready for review.

migueldeicaza:
Thank you, this makes sense.   U had initially missed the bit about the character drawing challenge, in that case, perhaps your color mixing is a better approach.   What do you think?

evertjr:
> Thank you, this makes sense. U had initially missed the bit about the character drawing challenge, in that case, perhaps your color mixing is a better approach. What do you think?

I think you're right, if we need to pre-composite for tiled characters anyway, doing it at the source for all dim text is simpler. I'll revert to the background-blend approach in dimmedColor() and drop the separate opaqueComposite method

evertjr:
@migueldeicaza did you had time to take look on this approach?

migueldeicaza:
Apologies, I went on vacation and have been slow at checking GitHub.   This looks great.

PR #481, opened 2026-02-27
Fix: Undefined reference to `initialize(fromContentsOf:)`  linker error on Linux

zxss702:
### Description
This PR fixes a critical linker error that prevents `SwiftTerm` from compiling successfully on Linux (e.g., Ubuntu) and Windows environments, especially when using newer Swift toolchains (like Swift 6.2).

### Motivation and Context
When building on Linux, the clang linker fails with the following error in `BufferLine.swift`:
`undefined reference to '$sSr22_FoundationCollectionsE10initialize14fromContentsOfSis5SliceVySryxGG_tF'`

This happens because `UnsafeMutableBufferPointer.initialize(fromContentsOf:)` currently fails to export the correct underlying binary symbols for array slices in the Swift Corelibs implementation for non-Apple platforms. 

### What Changed?
I introduced `#if os(Linux) || os(Windows)` compiler directives around the 4 occurrences of `initialize(fromContentsOf:)` in `BufferLine.swift` (`init`, `resize` for growing/shrinking, and `copyFrom`).
- **On Linux & Windows:** Safely initializes memory element-by-element using a `for` loop and `initializeElement(at:to:)` to bypass the missing symbol bug.
- **On Apple Platforms:** Preserves the original `initialize(fromContentsOf:)` implementation to maintain optimal memory copy performance.

### How Has This Been Tested?
Compiled and tested successfully on Ubuntu 22.04 with Swift 6.2. The linker errors are completely resolved and the buffer lines initialize correctly.

### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,0 +1,96 @@
+{
+  "originHash" : "b5ff689c86958c5a25a1a8946ddd3297f24a4ac1033e88edd10786f2c69a6415",
+  "pins" : [
+    {
+      "identity" : "hdrhistogram-swift",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/HdrHistogram/hdrhistogram-swift.git",
+      "state" : {
+        "revision" : "de0b9b8a27956b9bfc9b4dce7d1c38ad7c579f19",
+        "version" : "0.1.4"
+      }
+    },
+    {
+      "identity" : "package-benchmark",
+      "kind" : "remoteSourceControl",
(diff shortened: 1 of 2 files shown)


migueldeicaza:
This patch looks fine, I am sorry to hear that this regressed in the code and recent swift compilers.

Let me know when it is ready, the PR is flagged draft.

PR #487, opened 2026-03-12
don't percent encoding the url

hanyang-tony:
NSURL handles that itself, percent encoding whole url is not allowed.

from: https://developer.apple.com/documentation/foundation/nsstring/addingpercentencoding(withallowedcharacters:)

> Entire URL strings cannot be percent-encoded, because each URL component specifies a different set of allowed characters.

Encoding the url with query character set will cause url anchor to be invalid. For example, https://github.com/migueldeicaza/SwiftTerm/blob/main/README.md?plain=1#L11 becomes https://github.com/migueldeicaza/SwiftTerm/blob/main/README.md?plain=1%23L11


migueldeicaza:
Thank you!

PR #491, opened 2026-03-13
fix: WindowManipulationCommand crash and toSgr truecolor

jemdiggity:
## Summary

Two fixes:

1. **`WindowManipulationCommand` crash** — `case .resizeTo(_)` in `MacTerminalView.swift` only matches the single-argument `resizeTo(lines:)` variant. The two-argument `resizeTo(cols:rows:)` dispatched by `cmdWindowOptions` falls through with no match, causing: `Fatal error: unexpected enum case while switching on value of type 'WindowManipulationCommand'`.

   **Root cause:** Swift cannot disambiguate enum cases with the same base name but different arities in pattern matching ([SR-10077](https://github.com/apple/swift/issues/52479)). `case .resizeTo(_, _)` doesn't match `resizeTo(cols:rows:)` — it only matches the single-arg variant with a tuple decomposition.

   **Fix:** Added a new `resizeTerminal(cols:rows:)` case to `WindowManipulationCommand` as an additive, non-breaking change. Changed `cmdWindowOptions` to dispatch `.resizeTerminal` instead of `.resizeTo(cols:rows:)`. Added explicit handling in `MacTerminalView`'s switch. Existing API consumers are unaffected — `resizeTo(cols:rows:)` still exists but is no longer dispatched internally.

2. **`Attribute.toSgr()` debug spam** — Had `print("BROKEN")` statements instead of actually encoding truecolor values. Implemented proper SGR `38;2;r;g;b` and `48;2;r;g;b` sequences for truecolor foreground/background.

## Repro for crash

Run any program that sends `CSI 8 ; rows ; cols t` (xterm resize request) — e.g., neovim does this on startup. The crash is in `TerminalView.windowCommand` because the switch doesn't cover the two-argument `resizeTo(cols:rows:)` case.

## Approach

- **Additive, not breaking** — no existing enum cases renamed or removed
- New `resizeTerminal(cols: Int, rows: Int)` case added with documentation explaining why it exists
- Internal dispatch updated to use the new case
- Old `resizeTo(cols:rows:)` kept for backwards compatibility

## Test plan

- [x] `swift build --target SwiftTerm` compiles clean
- [x] Run neovim in a terminal — no longer crashes on window manipulation sequences
- [x] Truecolor SGR output verified (programs using 24-bit color)

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -1657,7 +1657,7 @@ extension TerminalView {
         userScrolling = false
     }
     
-    func scrollTo (row: Int, notifyAccessibility: Bool = true)
+    public func scrollTo (row: Int, notifyAccessibility: Bool = true)
     {
         let displayBuffer = terminal.displayBuffer
         if row != displayBuffer.yDisp {


migueldeicaza:
Thank you for both changes!

PR #497, opened 2026-03-21
Fix MacCaretView transparency support

Rostmen:
Add `layer?.isOpaque = false` to match the iOS implementation in iOSCaretView.swift. This enables proper cursor rendering when the terminal view has a transparent background.

Without this fix, the cursor becomes invisible when nativeBackgroundColor has alpha < 1.0 because CoreAnimation's layer compositing doesn't properly handle the CaretView's drawing.


migueldeicaza:

PR #498, opened 2026-03-23
Fix synchronized output rendering (DEC mode 2026)

inbarajaldrin:
## Summary

SwiftTerm's existing DEC 2026 implementation correctly snapshots the buffer via `displayBuffer`, but never suppresses rendering during sync blocks. `draw()` calls continue via `queuePendingDisplay`, `updateDisplay`, `scrollTo`, and `ensureCaretIsVisible` — causing visible scroll-through artifacts when applications send rapid screen updates through terminal multiplexers like tmux.

**Before:** Every sync block produces intermediate renders showing partially-repainted screens — visible as a rapid scroll-through from top to bottom.

**After:** Rendering is fully suppressed during sync blocks and the inter-block gaps. One atomic frame renders when the update sequence settles.

## The Problem

When running Claude Code (or similar TUI apps) inside tmux, tmux repaints the screen using multiple rapid BSU/ESU pairs delivered across separate I/O callbacks. Between these sync blocks, `queuePendingDisplay()` fires and triggers `draw()`, which renders partially-repainted intermediate states. This is visible as a scroll-through artifact that gets worse as the terminal session accumulates more content.

## The Fix

**Layer 1 — Sync block guards:** Suppress all rendering entry points while `synchronizedOutputActive` is true:
- `queuePendingDisplay()` — returns early
- `updateDisplay()` — returns early
- `scrollTo()` — returns early
- `ensureCaretIsVisible()` — returns early

**Layer 2 — Sync sequence coalescing:** Terminal multiplexers send multiple BSU/ESU pairs for a single logical screen update. An `inSyncSequence` flag stays true from the first BSU until `syncSequenceSettleMs` (default 100ms) after the last ESU. All rendering is suppressed during the entire sequence. One atomic render happens when the sequence fully settles. If a new BSU arrives within the window, the pending render is cancelled and the window resets.

The 100ms coalescing delay is the same pattern used by WezTerm (poll coalescing), xterm.js (requestAnimationFrame debounce), and Kitty (read batching). It only applies during sync sequences — normal terminal rendering is completely unaffected.

`synchronizedOutputActive` is made `public private(set)` so the view layer can read the sync state for guard checks.

## Changes

| File | Change |
|------|--------|
| `Terminal.swift` | `synchronizedOutputActive` → `public private(set)` |
| `AppleTerminalView.swift` | Guards on `queuePendingDisplay`, `updateDisplay`, `scrollTo`; `synchronizedOutputChanged` rewritten with sync sequence coalescing |
| `MacTerminalView.swift` | Guard on `ensureCaretIsVisible`; added `inSyncSequence`, `syncEndRenderTimer`, `syncSequenceSettleMs` properties |
| `iOSTerminalView.swift` | Same properties and `ensureCaretIsVisible` guard for platform parity |

## Test plan

- [x] Existing `SynchronizedOutputTests` pass (buffer snapshot behavior unchanged)
- [x] Tested with Claude Code inside tmux — scroll-through eliminated
- [x] Verified: normal typing, scrolling, copy-mode, Shift+Enter all unaffected
- [x] Verified: 1-second sync timeout safety valve still works
- [x] Build succeeds on macOS (xcodebuild)

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -343,6 +343,7 @@ extension TerminalView {
                 self.inSyncSequence = false
                 self.updateScroller()
                 self.queuePendingDisplay()
+                self.terminalDelegate?.scrolled(source: self, position: self.scrollPosition)
             }
             syncEndRenderTimer = work
             DispatchQueue.main.asyncAfter(
@@ -1872,8 +1873,6 @@ extension TerminalView {
     
     public func scrollTo (row: Int, notifyAccessibility: Bool = true)
     {
-        // Suppress during sync blocks and inter-block gaps.
-        guard !terminal.synchronizedOutputActive && !inSyncSequence else { return }
         let displayBuffer = terminal.displayBuffer
         if row != displayBuffer.yDisp {
             terminal.setViewYDisp (row)
--- a/Tests/SwiftTermTests/SynchronizedOutputTests.swift
+++ b/Tests/SwiftTermTests/SynchronizedOutputTests.swift
@@ -1,8 +1,11 @@
+import Foundation
 import Testing
 @testable import SwiftTerm
 
 final class SynchronizedOutputTests {
     private class TestDelegate: TerminalDelegate {
+        var scrolledPositions: [Int] = []
+
         func showCursor(source: Terminal) {}
         func hideCursor(source: Terminal) {}
         func setTerminalTitle(source: Terminal, title: String) {}
@@ -10,7 +13,9 @@ final class SynchronizedOutputTests {
         func windowCommand(source: Terminal, command: Terminal.WindowManipulationCommand) -> [UInt8]? { return nil }
         func sizeChanged(source: Terminal) {}
         func send(source: Terminal, data: ArraySlice<UInt8>) {}
-        func scrolled(source: Terminal, yDisp: Int) {}
+        func scrolled(source: Terminal, yDisp: Int) {
+            scrolledPositions.append(yDisp)
+        }
         func linefeed(source: Terminal) {}
         func bufferActivated(source: Terminal) {}
         func bell(source: Terminal) {}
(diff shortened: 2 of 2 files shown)


migueldeicaza:
Thanks for looking into this, the patch has a couple of problems:

# User scrollback is dropped during the sync window

The new guard in `scrollTo(row:notifyAccessibility:)` returns immediately whenever `terminal.synchronizedOutputActive || inSyncSequence`:

```swift
guard !terminal.synchronizedOutputActive && !inSyncSequence else { return }
```

That means wheel, scrollbar, and pan-driven viewport changes are ignored for the entire DEC 2026 session plus the debounce window. This is a regression in interactivity for long synchronized-output workloads.

This looks avoidable because `Terminal.setViewYDisp(_:)` already mirrors `yDisp` into both the live buffer and the synchronized snapshot:

```swift
buffer.yDisp = newValue
synchronizedOutputBuffer?.yDisp = newValue
```

So scrolling the frozen display buffer appears to be safe, and suppressing user scroll input here is broader than necessary.

# Final sync repaint does not notify external scroll observers

When the debounce timer fires after sync ends, the closure does this:

```swift
self.updateScroller()
self.queuePendingDisplay()
```

but it does not emit `terminalDelegate?.scrolled(...)`.

Normally the delegate gets scroll-position updates from the standard scroll callback paths (see the view files)

During synchronized output, those regular callbacks are based on the frozen `displayBuffer`, so host UI that depends on the delegate can stay stale after the deferred repaint completes.

We should Allow viewport changes during synchronized output if they only update `yDisp` against the frozen snapshot; 
Re-emit the scroll delegate notification when the deferred sync-end render is committed; Add view-level regression tests for those.

inbarajaldrin:
Thanks for the detailed review. I've addressed both issues:

**1. User scrollback no longer dropped during sync window**

Removed the blanket guard from `scrollTo(row:notifyAccessibility:)`. You're right that `setViewYDisp(_:)` already mirrors `yDisp` into both the live buffer and the synchronized snapshot, so user-initiated viewport changes are safe during sync.

`ensureCaretIsVisible()` retains the guard — that's the only path that needs suppression to prevent programmatic scroll-through.

**2. Scroll delegate notification after sync-end render**

The debounce timer closure now emits `terminalDelegate?.scrolled(source:position:)` after `updateScroller()` + `queuePendingDisplay()`, so host UI picks up the final scroll position.

**3. Regression tests added**

Added both model-level and view-level (`#if os(macOS)`) tests to `SynchronizedOutputTests.swift`:

- `testViewportScrollDuringSyncUpdatesBothBuffers` — verifies `setViewYDisp` updates both buffers during sync
- `testScrollDelegateFiredAfterSyncEnds` — verifies delegate notification after sync ends
- `testViewScrollToDuringSyncIsNotBlocked` — view-level: `TerminalView.scrollTo()` works during sync
- `testViewEmitsScrollDelegateAfterSyncEnd` — view-level: verifies state after debounce timer settles

migueldeicaza:
Thank you for addressing these issues!

PR #500, opened 2026-03-25
Fix: Kitty keyboard protocol sequences incorrectly trigger cursor save/restore

mauricehofman:
## Problem

`cmdRestoreCursor` (CSI `u`) and `cmdSaveCursor` (CSI `s`) do not check the `collect` (intermediate) parameter. This causes Kitty keyboard protocol sequences to be misinterpreted as cursor operations:

- `CSI > Ps u` (Kitty keyboard push) triggers cursor restore
- `CSI < u` (Kitty keyboard pop) triggers cursor restore  
- `CSI > Ps s` would trigger cursor save

Applications like [Claude Code](https://claude.ai/claude-code) send `ESC[>1u` and `ESC[<u` on every render cycle. Each occurrence incorrectly restores the cursor to a previously saved position, corrupting the display when the application uses relative cursor positioning (cursor up/down/forward).

## Fix

Add `guard collect.isEmpty else { return }` to both `cmdRestoreCursor` and `cmdSaveCursor`, matching the existing pattern in `cmdCsiM` which already correctly checks for the `>` intermediate.

## Affected applications

Any CLI tool using the Kitty keyboard protocol (progressive enhancement) will trigger this bug. This includes Claude Code 2.1.83+, which sends these sequences at startup and before each screen update.

### DIFF
--- a/Sources/SwiftTerm/EscapeSequenceParser.swift
+++ b/Sources/SwiftTerm/EscapeSequenceParser.swift
@@ -416,6 +416,11 @@ public class EscapeSequenceParser {
         case 0x71: terminal.cmdSetCursorStyle(pars, collect)    // q
         case 0x72: terminal.cmdSetScrollRegion(pars, collect)   // r
         case 0x73:                                              // s
+            // Only plain CSI s is overloaded between save-cursor and DECSLRM.
+            // Sequences with intermediates must not be routed to either handler.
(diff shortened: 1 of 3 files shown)


migueldeicaza:
Thanks, I also added a couple of additional scenarios that would have been wrong.

PR #501, opened 2026-03-25
Fix macOS dictation support in NSTextInputClient

SilverMarcs:
It would behave identical to Apple terminal app with these changes

Generated with Claude:

The NSTextInputClient implementation on macOS had stub/incomplete marked text methods that caused dictation to silently fail (system "bonk" sound) and IME composition to not track state properly.

Changes:
- Track marked text state (markedTextStorage/markedSelectedRange) across setMarkedText, unmarkText, and insertText
- selectedRange() now returns the cursor position as a zero-length range instead of NSRange(NSNotFound, 0) when there is no selection, giving the input system a valid anchor for dictation/IME
- hasMarkedText() and markedRange() return actual state instead of hardcoded false/empty
- attributedSubstring(forProposedRange:) returns marked text content when the requested range overlaps
- validAttributesForMarkedText() returns standard attributes
- characterIndex(for:) computes grid position from screen coordinates
- Add a floating overlay (NSTextField) at the caret position that shows live dictation/IME composition text with an underline, matching the terminal font and colors

### DIFF
--- a/Sources/SwiftTerm/Mac/MacTerminalView.swift
+++ b/Sources/SwiftTerm/Mac/MacTerminalView.swift
@@ -137,6 +137,11 @@ open class TerminalView: NSView, NSTextInputClient, NSUserInterfaceValidations, 
     var cellDimension: CellDimension!
     var caretView: CaretView!
     public var terminal: Terminal!
+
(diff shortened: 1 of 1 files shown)


migueldeicaza:

PR #502, opened 2026-03-25
Fix Home/End keys not sending escape sequences on macOS

colemaco:
## Summary

On macOS, pressing the Home/End keys triggers `scrollToBeginningOfDocument:` / `scrollToEndOfDocument:` via the NSResponder chain. `doCommand(by:)` only handled `moveToBeginningOfLine:` / `moveToEndOfLine:` (which map to Cmd+Left/Right), so Home/End fell through to the default case and were silently dropped with an "Unhandle selector" print.

This adds the missing cases in both the standard and Kitty keyboard protocol paths, sending the same escape sequences already used for `moveToBeginningOfLine:` / `moveToEndOfLine:`.

## Changes

- Handle `scrollToBeginningOfDocument:` → send Home escape sequence (`\e[H` / `\eOH`)
- Handle `scrollToEndOfDocument:` → send End escape sequence (`\e[F` / `\eOF`)
- Applied to both standard mode and Kitty keyboard protocol mode

## Test plan

- [x] Verified Home/End keys work at shell prompt (bash/zsh)
- [x] Verified Home/End keys work in vim (normal and insert mode)


migueldeicaza:

PR #503, opened 2026-03-26
Fix Kitty keyboard protocol arrow key mapping

nicolaregattieri:
## Summary
- Fix arrow keys being incorrectly mapped to keypad variants (57419/57420/57417/57418) due to macOS setting `.numericPad` on regular arrow key events
- Reset Kitty keyboard state on RMCUP (alternate screen exit) to prevent dirty state when apps like Claude Code exit without popping

## Context
Claude Code enables Kitty keyboard protocol. Without these fixes:
1. Arrow keys send `\e[57419u` (keypadUp) instead of `\e[A` (up), appearing as garbled text
2. After Claude exits, the terminal stays in Kitty mode, breaking the shell too


nicolaregattieri:
Opened on wrong repo by mistake, closing immediately.

migueldeicaza:
No problem, I do see that you have a fork with lots of changes, are some of those worth submitting to this version?

PR #505, opened 2026-03-27
Disable opportunistic Subprocess imports in LocalProcess

zxss702:
## Summary

This change disables the dormant `swift-subprocess` code path in `LocalProcess` by replacing all `#if canImport(Subprocess)` checks with `#if false //canImport(Subprocess)`.

## Why

`SwiftTerm` does not currently declare `Subprocess` as a target dependency, but `LocalProcess.swift` still contains conditional imports and execution paths for it.

In downstream projects, adding `swift-subprocess` at the app level can unexpectedly make `SwiftTerm` try to compile against `Subprocess`, even though `SwiftTerm` itself does not own that dependency. This can lead to build failures such as:

`missing required module '_SubprocessCShims'`

## What changed

- Disabled all `Subprocess`-guarded blocks in `LocalProcess`
- Kept `LocalProcess` on the existing `forkpty` implementation
- Prevented parent package dependencies from implicitly changing SwiftTerm's compile path

## Scope

This is an intentionally small stability fix. It does not introduce new behavior; it only prevents an undeclared optional dependency from affecting SwiftTerm builds.

## Validation

- `swift build --target SwiftTerm`

### DIFF
--- a/Sources/SwiftTerm/LocalProcess.swift
+++ b/Sources/SwiftTerm/LocalProcess.swift
@@ -9,13 +9,9 @@
 #if !os(iOS) && !os(Windows)
 import Foundation
 import Dispatch
-#if canImport(Subprocess)
+#if false //canImport(Subprocess)
 import Subprocess
-#if canImport(System)
 import System
-#else
-import SystemPackage
-#endif
 #endif
 
 /// Delegate that is invoked by the ``LocalProcess`` class in response to various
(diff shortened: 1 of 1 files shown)


zxss702:
Sorry, I seem to have messed up the git record. I don't know what I ordered.

zxss702:
But I'm sure it needs to be fix like this.

migueldeicaza:
Oh this is very interesting, I had no idea. 

Thanks for the patch!

Issue #482, opened 2026-02-27
canScroll and marginMode


holzschu:
In `Sources/SwiftTerm/Terminal.swift`, there is a variable, `canScroll`, defined as:
```swift
let canScroll = buffer.x >= buffer.marginLeft && buffer.x <= buffer.marginRight
```
But `marginLeft` and `marginRight` are only defined to something meaningful if `marginMode` is enabled (otherwise they're just 0 and 80, respectively).

Shouldn't this variable defined as:
```swift
let canScroll = !marginMode || (buffer.x >= buffer.marginLeft && buffer.x <= buffer.marginRight)
```
so that it defaults to `true` if `marginMode` is not enabled?

(This maybe a stupid question; I don't know what marginMode is and what the expected behaviour is in the terminal).

migueldeicaza:
I think it could

PR #507, opened 2026-03-29
The `canScroll` boolean defaults to `true` if `marginMode` is not set

holzschu:
This is related to issue #482 
Currently, `canScroll` takes into account the left and right margins set in the buffer, even if `marginMode` is not set. 

The PR adds `!marginMode ||` at the beginning of the test, so that `canScroll` defaults to `true` if `marginMode` is not set.

I identified this issue when the terminal outputs long lines (longer than 80 characters) and the next line begins over the previous line.

It also contains another fix: call `UInt8(min(255, something))` instead of `min(255, UInt8(something))` as the latter causes a crash when `something` is larger than 255.

(This is the first in a series of PR I'm making. I'm splitting my changes into small, separate PRs, so you can review them independently).


migueldeicaza:
This looks fine to me, thanks for waiting!

PR #508, opened 2026-03-29
Do not set terminal size (cols and rows) when called with a size of 0x0

holzschu:
Do not set terminal size (cols and rows) when called with a size of 0x0

Under some (unknown) circumstances, at the beginning of the app lifecycle, SwiftUI calls `init()` for the terminal with a size of 0 lines and 0 columns. There are fallbacks in the code for that case, and the minimum terminal size is 2x2. But this is not enough, as terminal commands can be executed with this size of 2x2 before the terminal is finally set to its correct size.

This PR detects the special case of 0 times 0, and does not set `rows` and `cols` until we get a proper size.

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -182,7 +182,7 @@ extension TerminalView {
             terminal.resize (cols: newCols, rows: newRows)
             
             // These used to be outside
-            // accessibility.invalidate ()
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Happy to merge it, but I wonder why I am not triggering that at all.

migueldeicaza:
Ouch, this broke the test suite.

migueldeicaza:
I think that we need an alternative mechanism to this.

PR #509, opened 2026-03-29
Adding VoiceOver support to SwiftTerm for iOS

holzschu:
This PR implements VoiceOver for iOS, following the instructions at https://developer.apple.com/videos/play/wwdc2019/248

Each line of the terminal is identified as its own reading area, and the textual content from the line is set as the Accessibility content. If it's activated, VoiceOver reads the lines as the user touches them.


migueldeicaza:
This is lovely, thank you so much!

PR #510, opened 2026-03-29
Enable Metal GPU rendering for SPM consumers

djensenius:
`canImport(MetalKit)` returns false when SwiftTerm is built as an SPM dependency because SPM doesn't add framework search paths for system frameworks. Fix by:
1. Define METAL_AVAILABLE via swiftSettings for iOS/macOS/visionOS
2. Change all #if canImport(MetalKit) to also check METAL_AVAILABLE
3. Add MetalKit/Metal linkerSettings for linking

Disclosure, this was done with AI assistance. It works for my project though.

### DIFF
--- a/Sources/SwiftTerm/Apple/Metal/MetalTerminalRenderer.swift
+++ b/Sources/SwiftTerm/Apple/Metal/MetalTerminalRenderer.swift
@@ -1822,26 +1822,21 @@ final class MetalTerminalRenderer: NSObject, MTKViewDelegate {
         rgb.getRed(&r, green: &g, blue: &b, alpha: &a)
         return SIMD4<Float>(Float(r), Float(g), Float(b), Float(a))
         #else
-        var r: CGFloat = 0
-        var g: CGFloat = 0
-        var b: CGFloat = 0
-        var a: CGFloat = 1
+        // Convert to sRGB via CGColor to handle all UIColor color spaces.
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Oh interesting, I was not aware of that.   I am very confused as to why the canImport(MetalKit) does not work, because I am using this on the sample MacTerminal, so this is a bit confusing.

Another option, that I think might work better is to use the new trait support, if you are on Xcode 26.4, this should now work.

djensenius:
I'll look into that! Also my branch has some other fork has some experiments sneaking in so I'm going to close this PR!

migueldeicaza:
Oh that looks interesting, is there some colorspace stuff I messed up?

djensenius:
@migueldeicaza I'm not sure, it appears so right now. It could also 100% be on me. I'm using it with a private (for now) terminal thing I'm working on. Mostly testing on iOS/VisionOS as opposed to macOS. I could send you a TestFlight or invite you to the private repo but building is complex and weird.

migueldeicaza:
That's fine, what I am mostly concerned about is getting the SPM package to work with Metal, and hoping we can find a solution.

djensenius:
Happy to work on that tonight ET or tomorrow. I'll see what I can do with trait support and how that works. I know the initial fix I proposed works, but trait management sounds better.

PR #511, opened 2026-03-30
Improving accessibility: VoiceOver for wrapped lines, speak the current page.

holzschu:
Building upon PR #509 and your work in this branch, these are suggestions for improvement:
- if a sentence goes over multiple lines (isWrapped, or automatic wrap in Vim), extend the selection area to get the full line. It affects `accessibilityContent(forLineNumber lineNumber: Int)`, `accessibilityFrame(forLineNumber lineNumber: Int)` and `accessibilityAttributedContent(forLineNumber lineNumber: Int)`.
- `accessibilityPageContent()`  and `accessibilityAttributedPageContent()` only speak the currently visible page.
- `accessibilityScroll()` deals with all possible options (up, down, left, right, next, previous).

There are still things I'm not satisfied with:
- text selection (but it's tricky).
- scrolling up or down requires three-fingers-swipe, I'm not sure how to change that.
- sometimes, the `isWrapped` boolean gets the wrong value (I did `cat sampleTextFileWithLongLines.md`, then `ls` in a big directory a couple of times, then `cat sampleTextFileWithLongLines.md`; the second time the long lines were not always grouped together. I tracked it down with the debugger and found that `isWrapped` was not set).

If you want to see it in action, it's available here: https://testflight.apple.com/join/WUdKe3f4

### DIFF
--- a/Sources/SwiftTerm/iOS/iOSTerminalView.swift
+++ b/Sources/SwiftTerm/iOS/iOSTerminalView.swift
@@ -2699,6 +2699,126 @@ extension TerminalViewDelegate {
 }
 
 extension TerminalView: UIAccessibilityReadingContent {
+    private func accessibilityBaseAttributes() -> [NSAttributedString.Key: Any] {
+        getAttributes(CharData.defaultAttr, withUrl: false) ?? [.font: fontSet.normal]
+    }
+
+    private func accessibilityAttributedLine(_ row: Int, endCol: Int = -1) -> NSAttributedString {
+        guard row >= 0, row < terminal.displayBuffer.lines.count else {
+            return NSAttributedString(string: "")
+        }
+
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Happy to merge this.

migueldeicaza:
Mhm there is a small conflict in the file, not sure how we have that.

holzschu:
I've made modifications to the functions you created, I should have see the conflicts coming. I've resolved the conflicts.
But I think it's good if you can review the code; I'm not sure about my use of `isWrapped` (it works, but should it work?)

migueldeicaza:
Your use of isWrapped looks fine to me, is there a particular thing that bothers you?

Since this is quite isolated, I am going to merge it.

PR #519, opened 2026-04-01
fix: forward mouse drag events in buttonEventTracking mode

### LINKED ISSUE #87: Mouse selection under tmux is broken
from Jason.

1. Make sure you have `set -g mouse on` inside ~/.tmux.conf (my full tmux config is here - https://github.com/nosami/configs/blob/master/.tmux.conf
2. Launch tmux session
3. Try and select text

same on XtermSharp.

gi11es:
## Summary

Fixes #87

- `mouseDragged` only forwarded drag events when `mouseMode` was `.anyEvent` (via `sendMotionEvent()`), but tmux uses `.buttonEventTracking` (`CSI ? 1002 h`) which also requires drag event forwarding for pane resize and mouse text selection.
- Changed the guard condition from `sendMotionEvent()` to `sendButtonTracking()`, which returns `true` for both `buttonEventTracking` and `anyEvent` modes.

## Reproduction

1. Open a tmux session inside SwiftTerm
2. Try to resize a pane by dragging the border
3. **Before fix:** nothing happens — drag events are silently dropped
4. **After fix:** pane resizes correctly, mouse selection works

## Test plan

- [x] Added `MouseTrackingTests` with 14 tests covering:
  - `MouseMode.sendButtonTracking()` returns correct values for all 5 enum cases
  - `MouseMode.sendMotionEvent()` returns correct values for all 5 enum cases
  - `sendButtonTracking()` is a superset of `sendMotionEvent()` (the invariant this fix relies on)
  - Integration: motion events produce SGR output when `buttonEventTracking` is active
  - Integration: motion events produce SGR output when `anyEvent` is active (regression guard)
  - Negative: `off` and `vt200` modes do not pass the `sendButtonTracking()` check
  - SGR encoding format matches what tmux expects (`<32;col;row M`)
- [x] All 14 tests pass via `swift test --filter MouseTrackingTests`

### DIFF
--- a/Sources/SwiftTerm/Mac/MacTerminalView.swift
+++ b/Sources/SwiftTerm/Mac/MacTerminalView.swift
@@ -2029,11 +2029,14 @@ open class TerminalView: NSView, NSTextInputClient, NSUserInterfaceValidations, 
         let mouseHit = calculateMouseHit(with: event)
         let hit = mouseHit.grid
         if allowMouseReporting {
-            if terminal.mouseMode.sendMotionEvent() {
+            // Forward drag events for both buttonEventTracking and anyEvent modes.
+            // buttonEventTracking sends motion events when a button is pressed,
+            // which is needed for tmux pane resize and mouse selection.
+            if terminal.mouseMode.sendButtonTracking() {
                 let flags = encodeMouseEvent(with: event)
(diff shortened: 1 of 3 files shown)


migueldeicaza:
Thank you!  We ended up with a different implementation bringing the change in in parallel, but I incorporated your test suite

PR #522, opened 2026-04-02
Fix: initialize buffer margins so CSI T scroll-down works on alt screen

colemaco:
Buffer.init never set _marginRight, leaving it at 0 instead of cols-1. cmdScrollDown (CSI T) unconditionally used marginRight to determine how many columns to copy per line, so on the alternate buffer it only copied 1 column, making scroll-down appear to update only the first lines.

This caused visible corruption in TUI apps like htop when scrolling up via mouse wheel, since htop uses CSI T to shift content down within its scroll region. reverseIndex (ESC M) was unaffected because it checks marginMode and moves whole line objects when margins aren't active.

Before screencap (notice the top ~10 processes are the only ones that refresh on scroll up):

https://github.com/user-attachments/assets/cf59b5ec-01a4-443c-968e-dc22cd57a3a3

### DIFF
--- a/Sources/SwiftTerm/Buffer.swift
+++ b/Sources/SwiftTerm/Buffer.swift
@@ -281,6 +281,8 @@ public final class Buffer {
         xBase = 0
         _scrollTop = 0
         _scrollBottom = rows - 1
+        _marginLeft = 0
+        _marginRight = cols - 1
         linesTop = 0
         _x = 0
         _y = 0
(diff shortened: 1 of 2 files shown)


migueldeicaza:
Thank you for your contribution, this looks correct, I will add tests.

saeedvaziry:
@migueldeicaza will there be a new release soon?

migueldeicaza:
Yes, I will cook one up soon

PR #530, opened 2026-04-07
feat: expose userScrolling for scroll-lock during output

k3rnelpan11c:
## Summary
- Make `Terminal.userScrolling` public so host apps can set it when the user scrolls away from the bottom
- Guard selection clearing in `feedPrepare()` and `linefeed()` with the `userScrolling` flag so active selections persist while new output streams in
- `scrollWheel` changed from `public` to `open` in `MacTerminalView` to allow subclass override

The `userScrolling` flag already exists and is checked in `Terminal.scroll()` (lines 5245, 5255, 5281) — these changes simply wire it up for external consumers.

## Motivation
When using terminal-based AI tools (like Claude Code) that produce long streaming output, users cannot:
1. Scroll up to read earlier output without being snapped back to the bottom
2. Select text without the selection being cleared by incoming data
3. Maintain a selection while new content arrives

These 3 minimal changes (~6 lines) fix all three issues by leveraging the existing `userScrolling` infrastructure.

## Test plan
- [ ] Verify default behavior unchanged (`userScrolling` defaults to `false`)
- [ ] Set `terminal.userScrolling = true` and verify scroll position is preserved during `feed()`
- [ ] Set `terminal.userScrolling = true` and verify text selection persists through `feed()`
- [ ] Verify `scrollWheel` can be overridden in subclass

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -1945,7 +1945,8 @@ extension TerminalView {
     {
         search.invalidate()
         // Preserve manual selection while output is streaming when mouse reporting is disabled.
-        if allowMouseReporting {
+        // Also preserve when user is scrolled back (viewing history while output arrives).
+        if allowMouseReporting && !terminal.userScrolling {
             selection.active = false
         }
         startDisplayUpdates()
--- a/Sources/SwiftTerm/Mac/MacTerminalView.swift
+++ b/Sources/SwiftTerm/Mac/MacTerminalView.swift
@@ -601,7 +601,8 @@ open class TerminalView: NSView, NSTextInputClient, NSUserInterfaceValidations, 
     
     open func linefeed(source: Terminal) {
         // Preserve manual selection while output is streaming when mouse reporting is disabled.
-        if allowMouseReporting {
+        // Also preserve when user is scrolled back (viewing history while output arrives).
+        if allowMouseReporting && !source.userScrolling {
             selection.selectNone()
         }
     }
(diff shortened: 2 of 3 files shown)


migueldeicaza:
Did you see the comments on #468?

k3rnelpan11c:
Thanks for the pointer to #468! I've reviewed your concerns there and here's how this PR addresses them:

**1. yDisp synchronization** — In our approach, `userScrolling` is set by the host app (Flock) from within the `scrollWheel` override, right after `super.scrollWheel()` has already updated yDisp. So the flag and yDisp stay in sync — we're not toggling the flag independently of scroll state.

**2. Buffer switching** — Good point. When the alternate screen buffer activates (vim, less, etc.), `canScroll` returns `false` and `scrollPosition` returns 0, so the host app's `scrollWheel` override naturally sets `userScrolling = false`. But we should verify this edge case more thoroughly.

**3. Delegate semantics** — Unlike #468, we don't hijack the delegate at all. The only change is making `userScrolling` public so the host app can set it. The existing `feedPrepare()` and `linefeed()` selection-clearing logic just gets an additional guard (`&& !terminal.userScrolling`). No delegate behavior changes.

**4. Scroll threshold** — `scrollWheel` fires on any scroll input, so there's no minimum threshold. The check is `scrollPosition >= 0.999` to detect "at bottom".

**5. Keyboard reset** — Addressed: the host app resets `userScrolling = false` in the `send()` override when the user types.

The changes are minimal (~6 lines in SwiftTerm) and all the coordination logic lives in the host app. Would you prefer we address anything else?

migueldeicaza:
Oh, I had not seen this update, will review.

migueldeicaza:
Would you mind testing this branch, and rebasing the above patch on this?  prevent-scroll-while-dragging

migueldeicaza:
I thin that the branch above takes care of it.  Closing.

PR #531, opened 2026-04-12
Add configurable fontSmoothing property

jimprosser:
## Summary

- Adds a public `fontSmoothing` property to `TerminalView` (defaults to `true`, preserving current behavior)
- When set to `false`, disables `CGContext.setShouldSmoothFonts` and `setAllowsFontSmoothing` on macOS, producing thinner text strokes on Retina displays — equivalent to iTerm2's "Thin strokes: On Retina Displays" setting
- Affects both the CoreText draw path (`AppleTerminalView`) and the Metal glyph rasterizer (`CoreTextGlyphRasterizer`)

## Motivation

Currently, font smoothing is hardcoded to `true` in both rendering paths. Terminal apps like iTerm2 offer this as a configurable option because many developers prefer thinner text strokes on high-DPI displays. Without this property, consumers of SwiftTerm cannot match that behavior without forking.

## Changes

- `AppleTerminalView.swift`: Add `fontSmoothing` property, use it in glyph drawing loop instead of hardcoded `true`
- `MacTerminalView.swift`: Add `_fontSmoothing` backing storage
- `CoreTextGlyphRasterizer.swift`: Add `fontSmoothing` property, use it in rasterize method
- `MetalTerminalRenderer.swift`: Sync `fontSmoothing` from `terminalView` to `rasterizer` each frame

## Usage

```swift
let terminalView = LocalProcessTerminalView(frame: .zero)
terminalView.fontSmoothing = false  // thin strokes on Retina
```

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -94,6 +94,17 @@ extension TerminalView {
         set { _fontSmoothing = newValue }
     }
 
+    /// Multiplier for vertical line spacing. 1.0 = default (ascent + descent + leading).
+    /// Set to 1.1 for 110% vertical spacing (matches iTerm2's vertical spacing setting).
+    /// Triggers a font reset and terminal resize when changed.
+    @objc open var lineSpacing: CGFloat {
+        get { _lineSpacing }
+        set {
+            _lineSpacing = newValue
+            resetFont()
+        }
+    }
+
     func resetCaches ()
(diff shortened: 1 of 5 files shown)


migueldeicaza:
I suspect the property setter needs to flag the view as needing an update, along with documentation.

There is an additional feature here, the scale, which I think we should remove and discuss separately

cheapsteak:
🙏 would much appreciate cutting a release for this

(almost forked and opened a duplicate pr XD)

PR #532, opened 2026-04-19
add Swift trait to enable/disable Subprocess

zadr:
Have this in a branch (along with a fork of TermKit to pick it up), but I'd like to get back on main. Seemed a bit nicer than hardcoding to `false` all over the place.


migueldeicaza:
I don't think that Subprocess will be fixed any time soon, I am debating removing it altogether.

zadr:
> I don't think that Subprocess will be fixed any time soon, I am debating removing it altogether.

that's fair. i'll close this pr out in that case 👋

PR #536, opened 2026-04-23
Support Shift+mouse to bypass application mouse reporting

evertjr:
When an application (e.g. tmux, vim, htop) enables mouse reporting, there is currently no way for the user to perform native text selection in the terminal. All mouse events are forwarded to the application.

The established convention across terminal emulators is to let users hold Shift to bypass mouse reporting and interact with the terminal's own selection instead. This PR adds that same behavior to SwiftTerm by checking for the Shift modifier in `mouseDown`, `mouseUp`, `mouseDragged`, and `scrollWheel` before forwarding events to the application. When Shift is held, the events fall through to normal selection/scroll handling.

### DIFF
--- a/Sources/SwiftTerm/EscapeSequenceParser.swift
+++ b/Sources/SwiftTerm/EscapeSequenceParser.swift
@@ -416,15 +416,14 @@ public class EscapeSequenceParser {
         case 0x71: terminal.cmdSetCursorStyle(pars, collect)    // q
         case 0x72: terminal.cmdSetScrollRegion(pars, collect)   // r
         case 0x73:                                              // s
-            // Only plain CSI s is overloaded between save-cursor and DECSLRM.
-            // Sequences with intermediates must not be routed to either handler.
-            if !collect.isEmpty {
-                break
-            }
-            if terminal.marginMode {
-                terminal.cmdSetMargins(pars, collect)
-            } else {
-                terminal.cmdSaveCursor(pars, collect)
+            if collect == [0x3E] {
+                terminal.cmdSetShiftEscape(pars)
+            } else if collect.isEmpty {
+                if terminal.marginMode {
+                    terminal.cmdSetMargins(pars, collect)
+                } else {
+                    terminal.cmdSaveCursor(pars, collect)
+                }
             }
         case 0x74: terminal.csit(pars, collect)                 // t
         case 0x75: terminal.cmdCsiU(pars, collect)              // u
--- a/Sources/SwiftTerm/Mac/MacTerminalView.swift
+++ b/Sources/SwiftTerm/Mac/MacTerminalView.swift
@@ -2062,7 +2062,7 @@ open class TerminalView: NSView, NSTextInputClient, NSUserInterfaceValidations, 
     }
     
     open override func mouseDown(with event: NSEvent) {
-        if allowMouseReporting && terminal.mouseMode.sendButtonPress() {
+        if allowMouseReporting && !event.modifierFlags.contains(.shift) && terminal.mouseMode.sendButtonPress() {
             sharedMouseEvent(with: event)
             return
         }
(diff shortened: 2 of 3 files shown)


migueldeicaza:
There is an idiom in the terminal app which is to flip a switch (on macOS that is command-r).

I’ll check what terminal.app is doing later to confirm; but I am worried some reporting modes would lose the shift encoding

evertjr:
Thanks for the feedback! I looked into how other terminals handle this, and the concern about losing the shift encoding is real.

**iTerm2** sidesteps it entirely by using Option (not Shift) as the bypass key, so shift is always encoded as +4 in mouse reports ([source](https://github.com/gnachman/iTerm2/blob/master/sources/TerminalView/PTYMouseHandler.m)). **Ghostty** takes a different approach: shift bypasses mouse reporting by default (matching xterm, Kitty, Alacritty, etc.), but it supports the `XTSHIFTESCAPE` escape sequence (`CSI > Ps s`) from xterm ([docs](https://ghostty.org/docs/vt/csi/xtshiftescape)). When a TUI app sends `CSI > 1 s`, shift stops being a bypass key and gets passed through as +4 in the mouse report instead. On exit, the app sends `CSI > 0 s` to restore the default. The sequence is defined in [xterm's control sequences spec](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html).

This gives users the behavior they expect out of the box while letting apps that genuinely need shift+click opt in to capturing it.

Would you be open to that approach here? The change would be:
- A `mouseShiftCapture` flag on `Terminal`, defaulting to off
- Handling `CSI > Ps s` to let apps toggle it
- The existing shift-bypass check becomes conditional on that flag

Happy to update the PR if that sounds reasonable.

migueldeicaza:
Thank you for the research, yes, I think that this is a good solution and would take that change.

We should still allow for the existing system that can be toggled on/off externally (in the Mac sample app, I bound it to the same command as MacOS terminal.app which is command-R) and on my own iOS app I think I have a button that does that toggling.

evertjr:
Updated the PR with XTSHIFTESCAPE support!

The shift bypass still works the same by default. The difference is that apps can now opt in to receiving shift+click by sending `CSI > 1 s`, and restore the default with `CSI > 0 s`. This is handled through a `mouseShiftCapture` flag on `Terminal` that the four mouse handlers check before deciding whether to bypass or forward.

The `allowMouseReporting` toggle stays as the master switch. When it's off, nothing gets reported regardless of shift or XTSHIFTESCAPE. So the Cmd+R binding in the Mac sample app and the iOS toggle button keep working exactly as before.

Matches xterm and [Ghostty's approach](https://ghostty.org/docs/vt/csi/xtshiftescape).

Sources/SwiftTerm/EscapeSequenceParser.swift
```diff
@@ -416,15 +416,14 @@ public class EscapeSequenceParser {
...
-            // Only plain CSI s is overloaded between save-cursor and DECSLRM.
-            // Sequences with intermediates must not be routed to either handler.
-            if !collect.isEmpty {
-                break
-            }
-            if terminal.marginMode {
-                terminal.cmdSetMargins(pars, collect)
-            } else {
-                terminal.cmdSaveCursor(pars, collect)
+            if collect == [0x3E] {
```
migueldeicaza, line 419: Let us use the standard idiom that I am trying to unify here: [UInt8(ascii: ">")]

Sources/SwiftTerm/EscapeSequenceParser.swift
```diff
@@ -416,15 +416,14 @@ public class EscapeSequenceParser {
         case 0x71: terminal.cmdSetCursorStyle(pars, collect)    // q
         case 0x72: terminal.cmdSetScrollRegion(pars, collect)   // r
         case 0x73:                                              // s
-            // Only plain CSI s is overloaded between save-cursor and DECSLRM.
```
migueldeicaza, line 419: This comment is removed, but the rationale below stayed, I think we should keep this comment.

migueldeicaza (approved):
It is also missing iOS support, but I think I can do it.

Luiginicoletti:
Amazing! @evertjr

PR #539, opened 2026-05-06
Fix Metal row cache stale content

evertjr:
## Problem

The Metal renderer can reuse a cached row even when the underlying `BufferLine` content changed in place. In alt-screen TUIs this can leave stale visual state on screen until another redraw happens.

I saw this with a long prompt where inserting a newline caused a blank line to visually disappear and glue adjacent rows together. Typing another character forced a repaint and healed the display.

Here some images to exemplify the issue:
metal-row-cache-before-shift-enter
I start with this, then add a new line

metal-row-cache-after-stale-line-glue
it incorrectly removes the empty line above text block and show this stale state until we type another letter.

## Fix

Track a lightweight fingerprint for each rendered row and include that fingerprint in the Metal row cache decision. Rows are now reused only when their rendered content is still the same.

## Tests

Manually verified in Maestri with the Metal renderer enabled using an alt-screen TUI and long wrapped input. Inserting a newline now updates immediately without stale glued rows.

### DIFF
--- a/Sources/SwiftTerm/Apple/Metal/MetalTerminalRenderer.swift
+++ b/Sources/SwiftTerm/Apple/Metal/MetalTerminalRenderer.swift
@@ -94,6 +94,7 @@ struct RowDrawBuffers {
 }
 
 struct RowCacheEntry {
+    var fingerprint: Int
     var data: RowDrawData?
     var buffers: RowDrawBuffers?
 }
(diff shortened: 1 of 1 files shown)


migueldeicaza:
This looks computationally very expensive (both the hash computation and this being called from Metal so often), I used a different approach that instead triggers a refresh in places that would make this go out of sync, can you please test it?

evertjr:
> This looks computationally very expensive (both the hash computation and this being called from Metal so often), I used a different approach that instead triggers a refresh in places that would make this go out of sync, can you please test it?

Tested and it fixes the original case! Your approach is better too, O(1) check, bumps at the mutation site, and lineRef === line catches CircularList rotations cleanly. Thanks for the quick turnaround.

PR #540, opened 2026-05-06
Pad Metal glyph atlas entries

evertjr:
## Problem

The Metal glyph atlas packs glyphs tightly together. With texture filtering, neighboring atlas entries can bleed into each other and produce visible rendering artifacts around text.

metal-glyph-atlas-claude-artifact

## Fix

Add padding around glyph atlas entries and duplicate edge pixels into that padding so sampled glyph edges do not pick up adjacent glyph data.

## Tests

Manually verified in Maestri with the Metal renderer enabled in text-heavy terminal UIs that previously showed blocky glyph artifacts.


migueldeicaza:
Ok, I merged your changes manually, but also avoided the slower code path for the colored path that was doing pixel at a time copies.

PR #541, opened 2026-05-06
Fix terminal background margin rendering

evertjr:
## Problem

When the terminal view width is not an exact multiple of the cell width, there is a small margin outside the grid on the right edge.

SwiftTerm was extending the last cell's background color into that margin. That makes application-controlled backgrounds visually protrude outside the terminal grid. If that extension is removed without filling the margin, the right edge can instead show an unpainted black strip.

right-margin-background-protrudes-full

## Fix

Keep application-set background colors inside the grid cells, and fill the non-grid margin with the terminal's default background color.

This matches the expected terminal behavior: applications control cell backgrounds, while the view margin belongs to the terminal background.

## Tests

Manually verified in Maestri with Gemini CLI in both the Core Graphics and Metal renderers. The input row no longer protrudes into the right margin, and the margin no longer appears as an unpainted strip.

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -1376,13 +1376,28 @@ extension TerminalView {
                             #endif
 
                             if endColumn >= terminal.cols {
-                                rect.size.width = frame.width - rect.origin.x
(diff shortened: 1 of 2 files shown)


migueldeicaza:

PR #542, opened 2026-05-06
Render local input responses immediately

evertjr:
## Problem

When DEC synchronized output mode is active, SwiftTerm can defer redraws while waiting for the sync sequence to settle. That is correct for application output, but it can make local interactive input feel stale when a TUI leaves synchronized output enabled longer than expected.

The typed bytes are processed, but the visual echo may appear delayed.

## Fix

Force an immediate redraw for local input responses so typing remains responsive even when synchronized output throttling is active.

The synchronized-output coalescing path is preserved for normal application output.

## Tests

Manually verified in Maestri with interactive terminal input while running TUIs that use synchronized output. Typed characters now render immediately without waiting for the settle fallback.

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -380,6 +380,14 @@ extension TerminalView {
                 self.queuePendingDisplay()
                 self.terminalDelegate?.scrolled(source: self, position: self.scrollPosition)
             }
+            guard syncSequenceSettleMs > 0 else {
+                if Thread.isMainThread {
+                    work.perform()
+                } else {
+                    DispatchQueue.main.async(execute: work)
+                }
+                return
+            }
             syncEndRenderTimer = work
             DispatchQueue.main.asyncAfter(
                 deadline: .now() + .milliseconds(syncSequenceSettleMs),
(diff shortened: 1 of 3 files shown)


migueldeicaza:
I am reviewing this one, there are a few edges cases that I am not quite sure are correct.

migueldeicaza:
I tried this, but also had a challenge that it was too slow before and after when using codex, so I figured I needed to take another look, can you check if the new branch on main fixes this for you?

evertjr:
Hey Miguel! I pulled the version from main and tested in Maestri. Your rework fixes the lag and it's cleaner. Thank you! Closing this.

evertjr:
Reopening with a tighter version on top of your rework.

After `9446f60`, the remaining typing-lag source is the 16.67ms throttle in `queuePendingDisplay`. When output bursts coincide with keystrokes, `pendingDisplay` stays armed and subsequent keystrokes get coalesced into the next 60Hz tick. Perceptible as occasional "skipped" characters that appear together a frame later.

This version is just the input-immediate bypass, no settle-window machinery. If output arrives within 150ms of a keystroke and we're not in a sync block, paint synchronously instead of via `asyncAfter(+16.67ms)`. Independent of and complementary to your sync-output simplification. Confirmed in Maestri against the same workloads that motivated the first PR.

PR #543, opened 2026-05-06
Fix hidden scroller layout on macOS

evertjr:
## Problem

On macOS, hiding SwiftTerm's built-in standalone scroller can still leave layout space reserved for it. Embedders that provide their own native scroller or no visible scroller can end up with an unnecessary blank strip on the right side of the terminal.

## Fix

Make the effective scroller width follow the scroller's visibility so hidden scrollers do not reserve layout width. Also ensure the scroller stays above the Metal view when it is visible.

## Tests

Manually verified in Maestri with both renderers. Hiding the built-in scroller no longer leaves a right-side gap, and the scroller is not covered when Metal is enabled at startup.


migueldeicaza:
Thank you for your contribution, this patch looks good.

PR #544, opened 2026-05-06
Report PTY pixel size from terminal cells

evertjr:
## Problem

`getWindowSize()` reports the view's raw frame size as `ws_xpixel`/`ws_ypixel`. When the view size includes fractional extra space outside the terminal grid, applications that inspect the PTY pixel size can see dimensions that do not match the reported row/column grid.

## Fix

Report PTY pixel dimensions from the actual terminal grid: `cols * cellWidth` and `rows * cellHeight`, converted to backing pixels.

This keeps the pixel size consistent with the row and column count.

## Tests

Manually verified in Maestri that local terminals continue resizing correctly after the change.


migueldeicaza:
This looks fine, thank you!

PR #545, opened 2026-05-06
Allow overriding Metal renderer scale

evertjr:
## Problem

The Metal renderer derives its scale from the view/window backing scale. Embedders that render terminal views inside transformed containers can need a different effective scale so glyphs remain sharp and correctly sized.

## Fix

Add an optional Metal scale factor override that embedders can set when they know the terminal is being rendered under an additional transform.

When unset, SwiftTerm keeps using the existing backing-scale behavior.

## Tests

Manually verified in Maestri with the Metal renderer inside a zoomable canvas. The default behavior remains unchanged when no override is set.


evertjr:
Sorry for the several PRs all of the sudden. I spend the last few days working on rendering improvements for my app and I think some of the fixes belong upstream too.

migueldeicaza:
Thank you so much, apologies for the delay in reviewing the changes, I am processing them little by little.

And I deeply appreciate upstreaming these changes, they will help everyone else as well.

PR #546, opened 2026-05-11
Tag CAMetalLayer with sRGB colorspace to match CoreGraphics rendering

evertjr:
## Problem

A regular layer-backed `NSView`/`UIView` writes its `draw(_:)` output to a backing store whose colorspace is set to the display colorspace, so the compositor color-manages those pixels, sRGB content gets converted to the display's gamut (sRGB or Display P3) at scan-out.

`CAMetalLayer.colorspace` defaults to `nil`, which tells the compositor to treat the layer's bytes as already-in-display-gamut and skip color management. On a wide-gamut display, `NSColor` values resolved via `usingColorSpace(.deviceRGB)`, which are sRGB-encoded on modern macOS, end up displayed as if they were Display P3 values, producing oversaturated, harsh-looking colors.

Most visible on selection highlights and any non-default cell backgrounds: the GPU-rendered selection looks solid and saturated while the CG-rendered selection looks subdued and translucent, even though both pipelines are fed the same source color.

I noticed this in my app (Maestri) where users could toggle Metal on/off and the selection color shifted noticeably between the two.

cg-selection-correct
Non metal text selection

metal-selection-broken
Metal text selection

## Fix

Tag the metal layer with `CGColorSpace.sRGB` after creating the `MTKView` (both macOS and iOS paths) so the compositor applies the same conversion it applies to the layer-backed `NSView`/`UIView` backing store. Same source color, same on-screen pixel through both renderers.

## Tests

Manually verified in Maestri on a wide-gamut display. Selection highlights now match between the Metal and CoreGraphics paths.

### DIFF
--- a/Sources/SwiftTerm/Mac/MacTerminalView.swift
+++ b/Sources/SwiftTerm/Mac/MacTerminalView.swift
@@ -304,6 +304,18 @@ open class TerminalView: NSView, NSTextInputClient, NSUserInterfaceValidations, 
             mtkView.autoResizeDrawable = false
             mtkView.framebufferOnly = true
             mtkView.colorPixelFormat = .bgra8Unorm
+            // Tag the metal layer with sRGB so the compositor color-manages our
+            // pixels the same way it color-manages the layer-backed NSView
+            // (whose backing store is in the display colorspace). Without this,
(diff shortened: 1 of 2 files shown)


migueldeicaza:
Nice find!

PR #547, opened 2026-05-11
Fix stale Metal cursor on cursor-only buffer moves

evertjr:
## Problem

Sequences like `CSI Ps C` / `CSI Ps D`  which shells emit in response to Option+Arrow word jumps and similar — mutate `buffer.x` without dirtying any row. `getUpdateRange()` returns `nil` and `updateDisplay` early-returns, so the Metal renderer is never asked to redraw. Because `MTKView` is paused and the Metal renderer draws the cursor itself (`caretView` is hidden in Metal mode), the cursor visually freezes at its old screen position until something else triggers a frame, typing a printable character, scrolling, etc.

Most visible when holding Option+Arrow to word-jump through a long shell prompt.

The bug is partially masked upstream by the caret blink timer firing `setNeedsDisplay` every 700ms, which incidentally catches up the stale cursor position. With a non-blinking cursor style (e.g. `steadyBlock`) the freeze is permanent until something else triggers a frame.

## Fix

Track the last-rendered cursor position in `TerminalView` and request a Metal redraw when it changes, even if no rows are dirty. The comparison is a 3-field tuple equality, cheap, and avoids redundant frames when nothing actually changed.

## Tests

Manually verified in Maestri with Metal enabled and a steady cursor. Holding Option+Arrow through a long shell prompt now tracks the cursor live instead of freezing it until the next keystroke.

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -1630,6 +1630,22 @@ extension TerminalView {
                 let y = buffer.yDisp+buffer.y
                 terminalDelegate?.rangeChanged (source: self, startY: y, endY: y)
             }
+            // Pure cursor moves (e.g. CSI C / CSI D from word-jumps) don't
+            // mark any row dirty, so getUpdateRange() returns nil. With Metal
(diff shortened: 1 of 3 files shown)


migueldeicaza:
Thank you!

PR #548, opened 2026-05-11
Rebind Metal renderer when TerminalView is reparented to a new window

evertjr:
> **Note**: Overlaps with #546 (sRGB colorspace tag). The `makeMetalView` helper introduced here includes that tag, so the two PRs touch the same setup code. Merging #546 first lets this one rebase trivially. Either merge order preserves the sRGB behavior, since the helper is the new single source of truth for MTKView configuration.

## Problem

If a `TerminalView` is moved between `NSWindow` instances (e.g. a SwiftUI host that sets `applicationShouldTerminateAfterLastWindowClosed = false` and rebuilds its `WindowGroup` window on dock-icon reopen), the Metal-rendered terminal comes back frozen on its last frame. PTY output keeps flowing, typed input reaches the shell, but nothing renders.

`CAMetalLayer` binds to its window's WindowServer surface on first attach and Apple exposes no public API to refresh that binding. After the reparent the layer keeps minting drawables and `commandBuffer.commit()` succeeds, but `present(_:)` silently no-ops against the stale binding. `MTKView.releaseDrawables()` doesn't help. The only reliable fix is to recreate the `MTKView` so a fresh layer binds to the new window.

I hit this in Maestri (terminals survive the main window being closed-to-dock), but it applies to any SwiftTerm host that re-parents.

https://github.com/user-attachments/assets/08d30ad0-747f-4095-b965-3614d69ad507

## Fix

Track which window the current `metalView` is bound to via a `weak var metalBoundWindow`. In `viewDidMoveToWindow`, if `window !== metalBoundWindow`, rebuild the MTKView.

The naive disable-then-re-enable causes a one-frame flash because the hierarchy briefly contains no Metal view. The rebind path instead builds the new view and renderer, inserts in the old view's z-position, forces a synchronous first frame, then detaches the stale view, so there's never a frame where CoreAnimation can commit the CoreGraphics fallback.

Two private helpers (`makeMetalView`, `insertMetalView`) are extracted so the initial-enable path and the rebind path share their MTKView configuration and z-order logic and can't drift apart.

On renderer init failure during rebind, fall back to CoreGraphics rather than leaving a frozen view in place. Errors go through `os_log`.

## Tests

Manually verified in Maestri with Metal enabled. Closing the window to dock and reopening now returns to a live terminal with no flash. Repeated close/reopen cycles work. Typing during the brief reparent doesn't lose input.

### DIFF
--- a/Sources/SwiftTerm/Mac/MacTerminalView.swift
+++ b/Sources/SwiftTerm/Mac/MacTerminalView.swift
@@ -17,6 +17,9 @@ import Carbon.HIToolbox
 #if canImport(MetalKit)
 import MetalKit
 #endif
+#if canImport(os)
+import os.log
+#endif
 
 /**
  * TerminalView provides an AppKit front-end to the `Terminal` termininal emulator.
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Thank you for your contribution!   Apologies for taking this long to review.

PR #553, opened 2026-05-19
Render local input responses immediately

evertjr:
Reopened version of #542 on top of the recent sync-output rework.

## Problem

After `9446f60`, the remaining typing-lag source is the 16.67ms throttle in `queuePendingDisplay`. When output bursts coincide with keystrokes, `pendingDisplay` stays armed and subsequent keystrokes get coalesced into the next 60Hz tick. Perceptible as occasional "skipped" characters that appear together a frame later.

In practice this surfaces most when the host is under load — we see it consistently when many terminals are running concurrently in Maestri.

## Fix

Input-immediate bypass only, no settle-window machinery. If output arrives within 150ms of a keystroke and we're not in a sync block, paint synchronously instead of via `asyncAfter(+16.67ms)`. Complementary to the sync-output simplification on main.

## Tests

Confirmed in Maestri against the same workloads that motivated the first PR.

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -355,7 +355,6 @@ extension TerminalView {
 
     public func synchronizedOutputChanged (source: Terminal, active: Bool)
     {
-        SyncDebug.log("delegate active=\(active)")
         if !active {
             updateScroller()
             queuePendingDisplay()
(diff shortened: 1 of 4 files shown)


evertjr:
Added a second commit dropping the `SyncDebug.log(...)` call sites in `Terminal.swift` and `AppleTerminalView.swift`. The `SyncDebug` type isn't defined anywhere in the tree, so main itself doesn't compile and CI on this PR was inheriting that failure. Easy to revert if you bring `SyncDebug` back.

migueldeicaza:
Apologies for the delay, question about this.

displayImmediately() can schedule a separate main-thread redraw for every feed chunk.

It looks like feedFinish() calls displayImmediately() without using pendingDisplay to coalesce redraws. Then when we run the feed on the main thread we post a new call to the main queue every time.

Seems like  the immediate path should still bypass the 16.67 ms timer, but it needs to coalesce duplicate redraw requests. For example, set pendingDisplay before posting the immediate main-queue block, then clear it once the redraw runs.

migueldeicaza:
Let me merge and add the suggestion

PR #558, opened 2026-05-24
Add selectedTextForegroundColor to recolor selected text

akingundogdu:
## Summary

Adds an optional `selectedTextForegroundColor` property to the macOS (`TerminalView`/`MacTerminalView`) and iOS terminal views.

When set, selected text is rendered in this color in addition to the existing `selectedTextBackgroundColor`. When `nil` (the default), behavior is unchanged — selected text keeps its original foreground color and only the selection background is drawn, so this is fully backward compatible and opt-in.

## Motivation

Several terminal color schemes define a distinct *selection foreground* alongside the selection background (for example GitHub Dark uses near-black text on a yellow selection). Today only the background can be customized, so selected text can become hard to read against a saturated selection background. This adds parity with terminals/themes that expect a selection foreground.

## Changes

- `Mac/MacTerminalView.swift`: add `selectedTextForegroundColor: NSColor?` (default `nil`).
- `iOS/iOSTerminalView.swift`: add `selectedTextForegroundColor: UIColor?` (default `nil`).
- `Apple/AppleTerminalView.swift`: in `buildAttributedString`, when a cell is selected, override the foreground attribute with `selectedTextForegroundColor` if it is set (the existing draw path already honors `.foregroundColor`).

## Testing

- `swift build --target SwiftTerm` succeeds.
- Verified in an embedding app: setting `selectedTextForegroundColor` renders legible dark text on a light selection background; leaving it `nil` preserves existing behavior.

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -645,6 +645,9 @@ extension TerminalView {
             if isSelected {
                 var mutable = attributes
                 mutable[.selectionBackgroundColor] = selectedTextBackgroundColor
+                if let selectedTextForegroundColor {
+                    mutable[.foregroundColor] = selectedTextForegroundColor
+                }
                 currentAttributes = mutable
             } else {
                 currentAttributes = attributes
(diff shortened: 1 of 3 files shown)


migueldeicaza:
This I believe is superseded by commit ed395c7. Let me know if it works for you

PR #560, opened 2026-05-30
Fix auto-scroll and selection during active output

SilverMarcs:
Sets terminal.userScrolling when the user scrolls away from the bottom so incoming output no longer yanks the viewport back down. Also stops clearing selection on every feed/linefeed unless the hosted app has actually requested mouse reporting (mouseMode != .off).


migueldeicaza:
I believe this is now fixed, can you confirm that this works for you?

SilverMarcs:
Works well now

PR #561, opened 2026-05-30
Improve macOS dictation and IME marked-text overlays

SilverMarcs:
Implements wrapped and exclusion-path bounded NSTextView rendering for dictation overlays (DictationOverlayTextView). This wraps long marked text input and paints background fills precisely under active text fragments per line-fragment to avoid clipping or overlapping prompt elements.

### DIFF
--- a/Sources/SwiftTerm/Mac/MacTerminalView.swift
+++ b/Sources/SwiftTerm/Mac/MacTerminalView.swift
@@ -179,7 +179,7 @@ open class TerminalView: NSView, NSTextInputClient, NSUserInterfaceValidations, 
     /// Marked (uncommitted) text from an input source (IME, dictation, etc.).
     private var markedTextStorage: NSAttributedString?
     private var markedSelectedRange: NSRange = NSRange(location: NSNotFound, length: 0)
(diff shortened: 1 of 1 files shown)


migueldeicaza:
This looks fabulous, how can I test this and take it for a spin?

SilverMarcs:
> This looks fabulous, how can I test this and take it for a spin?

You can just trigger the system wide dictation (f5 key) on Mac keyboards) and on a less wide window, the preview should overflow a lot better

migueldeicaza:
Ok, I tested it, I do like the new behavior!

One thing though, the color picking seems peculiar, and I am not sure what to do about ti, the sample default Mac app looks like black text being rendered on top of a very dark background, so it is barely noticeable.

SilverMarcs:
I think we can grab the effective background color of the terminal as well as the foreground color and use that as the colors of the overlay and its text. 
I have to explored the app extensively but I’m sure the package already has convenient ways of getting these properties 

What do you think?

migueldeicaza:
Ok, I merged this, and as a follow up I updated the color.

PR #564, opened 2026-06-03
Fix Option-composed characters dropped under the kitty keyboard protocol

evertjr:
## Problem

When the kitty keyboard protocol is active and a printable character is produced via Option acting as a compose/AltGr layer, the composed character can be silently dropped and the base-layout key is sent instead.

The clearest repro is a non-US layout where the composed glyph sits on a key whose unmodified character differs from US. On the **Czech** layout the number-row "2" key produces `ě` when pressed bare and `@` with Option. Typing **Option+2** in an app that enabled the protocol with `reportAlternates` (but not `reportText`) inserts `ě` instead of `@`.

The cause is in `insertText`. An Option-composed commit is turned into a kitty key event via `kittyTextEvent`, which keys the event off `charactersIgnoringModifiers` (the base-layout glyph, `ě`) and carries the actual `@` only as associated `text`. The encoder then:

- with `reportAlternates`, sees a non-nil base-layout alternate and bails out of the plain-text fast path in `KittyKeyboardEncoder.encode`, emitting a CSI-u sequence keyed on `ě`;
- only includes the associated `@` text when `reportText` (flag 16) is negotiated.

So any app that turns on `reportAlternates`/`reportAllKeys` without `reportText` receives the base key and never sees the composed character. OpenCode (Bubble Tea) is one such app, which is what surfaced this. US-layout Option characters (`Option+8` = `•`, etc.) are unaffected because their base glyph matches US, so the alternate is nil and they take the plain-text path. Apps that don't enable the protocol at all (e.g. Claude Code's default mode) are also unaffected, which is why this only reproduced in some terminals/apps.

## Fix

When `optionAsMetaKey` is off and Option produced a plain printable character with no Control/Command modifier, send the composed text directly as UTF-8 instead of routing it through the kitty encoder. This mirrors the existing non-report-all plain-text fast path and matches how kitty and Ghostty treat AltGr layers: the character the user typed is what reaches the program. It is gated on `!optionAsMetaKey`, so it is a no-op when Option-as-Meta is enabled (the default).

## Tests

Added `KittyOptionComposeTests.swift`, a view-level test that drives the Czech `Option+2` case (`charactersIgnoringModifiers = "ě"`, composed `@`) with the kitty alternates/all-keys flags pushed and `optionAsMetaKey` off, and asserts the PTY receives `@`. Without the fix the same input emits the base-layout CSI-u sequence instead.

Also verified live: switching to the Czech input source and typing `Option+2` in OpenCode produced `ě` before the fix and `@` after.

### DIFF
--- a/Sources/SwiftTerm/Mac/MacTerminalView.swift
+++ b/Sources/SwiftTerm/Mac/MacTerminalView.swift
@@ -1376,6 +1376,28 @@ open class TerminalView: NSView, NSTextInputClient, NSUserInterfaceValidations, 
                 pendingKittyKeyEvent = nil
                 kittyIsComposing = false
                 let text = str as String
+
+                // Option acting as a compose/AltGr layer (e.g. on the Czech
+                // layout Option+2 produces "@" while the bare key is "ě"). When
+                // optionAsMetaKey is off, Option is not Meta, so the keypress
+                // simply produced a normal character. Encoding it as a kitty
+                // CSI-u event keys off charactersIgnoringModifiers (the
+                // base-layout glyph, "ě") and only carries the composed "@" as
+                // associated text, so apps that negotiate reportAlternates or
+                // reportAllKeys without reportText (e.g. OpenCode) receive the
+                // base key and the composed glyph is lost. Send the composed
+                // text directly instead, matching how kitty/Ghostty handle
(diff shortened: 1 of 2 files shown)


migueldeicaza:
Thank you for the research and fix!

PR #567, opened 2026-06-06
fix(iOS): mark TerminalView non-opaque to stop scrollback rendering corruption

### LINKED ISSUE #566: iOS: scrollback renders garbled/striped (flickers) — TerminalView is never marked non-opaque
### Summary

On iOS, the scrollback renders as flickering, striped/garbled noise over every region that uses the **default background color**. Cells that have an explicit (non-default) background color render correctly. It is most obvious while scrolling, but with enough content it is visible at rest too.

| Before (1.13.0 / `main`) | After (`isOpaque = false`) |
|---|---|
| before – striped garbage over default-bg areas | after – clean |

Both screenshots feed the **same** synthetic content (iPhone 16 simulator, iOS 18.2). The only difference is the one-line fix below.

### Environment

- SwiftTerm **1.13.0** (also reproduces on current `main`)
- iOS **18.2** (iPhone 16 simulator; also observed on device)
- Default CoreGraphics renderer (Metal disabled — the default)

### Root cause

`TerminalView` (a `UIScrollView`) provides its background through `layer.backgroundColor` and paints each glyph cell with a **transparent** backdrop in `draw(_:)`. `setupOptions()` does:

```swift
layer.backgroundColor = nativeBackgroundColor.cgColor
nativeBackgroundColor = UIColor.clear   // cells are now filled with .clear
```

and `draw(_:)` then does `nativeBackgroundColor.set(); context.fill([dirtyRect])` (i.e. fills with `.clear`), expecting the layer colour to show through the gaps.

This scheme requires a **non-opaque** view, but `isOpaque` is never set to `false`, so it keeps the default `true`. An opaque view is given an **alpha-less** CoreGraphics context, so:

- the `UIColor.clear` fill in `draw(_:)` is a no-op, and
- every default-background region is left as **uninitialised backing-store memory**.

A full-screen repaint masks this, but `UIScrollView` optimises scrolling by blitting the existing content and only calling `draw(_:)` for the newly exposed strips — which exposes that uninitialised memory as the flickering, striped corruption above.

### Reproduction

A ready-to-run reproduction (no SSH/network) is on this branch:
**https://github.com/ferponse/SwiftTerm/tree/repro/ios-scroll-corruption**

It replaces the bundled `iOSTerminal` sample's root view controller with a self-contained synthetic feeder. Steps:

1. Check out `repro/ios-scroll-corruption` (branched off `main`).
2. Open `TerminalApp/iOSTerminal.xcodeproj`, run the **iOSTerminal** scheme on an iOS simulator.
3. The default-background areas render as striped garbage; scrolling the scrollback makes it flicker.

The essence of the repro (feed several screens, including default-bg lines, then scroll):

```swift
let tv = TerminalView(frame: view.bounds,
                      font: .monospacedSystemFont(ofSize: 13, weight: .regular))
// … add full-screen …
var s = ""
for i in 0..<60 {
    s += "bash-5.2# echo line \(i)\r\n"
    s += "Welcome — iteration \(i)\r\n"                 // default background
    var rainbow = ""
    for c in 0..<12 { rainbow += "\u{1b}[48;5;\(16 + c*18)m  \u{1b}[0m" }  // explicit bg
    s += rainbow + "\r\n"
}
tv.feed(text: s)   // scroll the scrollback → striped corruption over the default-bg lines
```

### Fix

Mark the view non-opaque so the transparent compositing behaves as designed:

```swift
// in setupOptions()
isOpaque = false
```

PR: #567 (verified on the iPhone 16 / iOS 18.2 simulator — clean at rest and while scrolling in both directions).

ferponse:
### Problem

On iOS, scrollback renders as flickering, striped garbage over default-background regions once the user scrolls (see #566).

| Before (1.13.0 / `main`) | After (this PR) |
|---|---|
| before – striped garbage | after – clean |

### Cause

`TerminalView` (a `UIScrollView`) draws its background via `layer.backgroundColor` and fills glyph cells transparently in `draw(_:)` (`setupOptions()` sets `nativeBackgroundColor = .clear`), relying on the layer colour showing through the gaps.

That requires a **non-opaque** view, but `isOpaque` was never cleared, so the view's CoreGraphics context had no alpha channel: the `UIColor.clear` fill in `draw(_:)` became a no-op and every default-background region was left as uninitialised backing-store memory. It is masked on full repaints and exposed by the scroll view's blit / redraw-only-the-exposed-strip behaviour during scrolling.

### Fix

Set `isOpaque = false` in `setupOptions()`. One line; it only enables the transparent compositing the existing code already assumes.

### Testing

Reproduced and verified on the iPhone 16 / iOS 18.2 simulator using the bundled `iOSTerminal` demo fed with synthetic coloured scrollback. A ready-to-run reproduction is on [`repro/ios-scroll-corruption`](https://github.com/ferponse/SwiftTerm/tree/repro/ios-scroll-corruption):

- **Before:** full-screen striped corruption over default-background regions (matches #566).
- **After:** clean background at rest and while scrolling in both directions; coloured cells and text unaffected.

Fixes #566.

### DIFF
--- a/Sources/SwiftTerm/iOS/iOSTerminalView.swift
+++ b/Sources/SwiftTerm/iOS/iOSTerminalView.swift
@@ -1179,6 +1179,16 @@ open class TerminalView: UIScrollView, UITextInputTraits, UIKeyInput, UIScrollVi
         setupOptions(width: bounds.width, height: bounds.height)
         layer.backgroundColor = nativeBackgroundColor.cgColor
         nativeBackgroundColor = UIColor.clear
+        // The terminal background is provided by `layer.backgroundColor`, and
+        // `draw(_:)` paints glyph cells with a transparent backdrop so that the
+        // layer colour shows through the gaps. That only works if the view is
+        // non-opaque: an opaque view gets an alpha-less graphics context where
+        // the transparent fill is a no-op, leaving uninitialised backing-store
+        // garbage in every default-background region. When the scroll view blits
+        // and re-exposes strips during scrolling, that garbage becomes visible as
+        // flickering/striped corruption. Marking the view non-opaque makes the
+        // transparent compositing behave as intended.
+        isOpaque = false
     }
     
     var _nativeFg, _nativeBg: TTColor!


migueldeicaza:
Thank yoU!

PR #571, opened 2026-06-09
Fix DECRST 1005/1006/1015/1016 disabling mouse tracking

AlexDemzz:
## Problem

The DECRST handlers for the mouse coordinate **encoding** modes (`?1005l`, `?1006l`, `?1015l`, `?1016l`) reset `mouseProtocol` (correct) but also set `mouseMode = .off`, turning mouse **tracking** off entirely.

Per xterm semantics the encoding modes are independent of the tracking modes (`?9`/`?1000`–`?1003`): resetting an encoding only reverts how coordinates are encoded, it must not stop tracking. The DECSET side already reflects this (`?1006h` only sets `mouseProtocol = .sgr` and doesn't touch `mouseMode`), so set/reset were asymmetric.

## Real-world impact

mosh re-asserts terminal state on every resize/reattach repaint as:

```
CSI ?1003l ?1003h ?1004l ?1006l ?1006h
```

With the current handler the trailing `?1006l` disables tracking right after `?1003h` re-enabled it, and `?1006h` only restores the encoding — so any app using mouse reporting over mosh (e.g. Claude Code, tmux with mouse on) permanently loses mouse input after the first terminal resize. Reproduced and verified against mosh 1.4.0 with a minimal client: `printf '\033[?1003h\033[?1006h'` + resize.

## Fix

DECRST 1005/1006/1015/1016 now only reset `mouseProtocol`, leaving `mouseMode` unchanged.

### DIFF
--- a/Sources/SwiftTerm/Terminal.swift
+++ b/Sources/SwiftTerm/Terminal.swift
@@ -4167,17 +4167,18 @@ open class Terminal {
             case 1004: // send focusin/focusout events
                 sendFocus = false
             case 1005: // utf8 ext mode mouse
+                // Resets the coordinate encoding only: 1005/1006/1015/1016 select how mouse
+                // events are encoded, independent of the tracking modes (9/1000-1003). xterm
+                // keeps tracking enabled when an encoding is reset; turning mouseMode off here
+                // broke clients (e.g. mosh re-asserts modes as "CSI ?1003l ?1003h ?1006l ?1006h"
+                // on every resize, which left tracking permanently disabled).
                 mouseProtocol = .x10
-                mouseMode = .off
             case 1006: // sgr ext mode mouse
                 mouseProtocol = .x10
-                mouseMode = .off
             case 1015: // urxvt ext mode mouse
                 mouseProtocol = .x10
-                mouseMode = .off
             case 1016: // sgrPixel mode
                 mouseProtocol = .x10
-                mouseMode = .off
             case 25: // hide cursor
                 hideCursor ()
             case 1048: // alt screen cursor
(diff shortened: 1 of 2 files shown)


migueldeicaza:
Thanks for the patch, are there tests cases I can use to validate this?

I will look at the docs, and compare notes, but pointers to cases would help me

AlexDemzz:
I've added test cases to the PR (in `MouseTrackingTests.swift`, ff59b8a) — all green on the branch, and the two tracking ones fail on current main with `mouseMode → .off`:

- **`decrstEncodingModeKeepsTrackingEnabled`** — the core asymmetry: for each of 1005/1006/1015/1016, feed `?1003h ?<mode>h ?<mode>l` and assert `mouseMode` is still `.anyEvent`.
- **`decrstEncodingModeStillResetsEncoding`** — the reset still does its actual job: after `?1006l`, a scroll event is reported in the default X10 encoding again (`ESC [ M ...` bytes).
- **`decrstEncodingModeReportedViaDecrqm`** — observable through DECRQM: after `?1006l`, `CSI ?1006$p` reports reset (`;2$y`) while `CSI ?1003$p` still reports set (`;1$y`).
- **`moshStyleModeReassertKeepsTrackingAndSgrEncoding`** — the real-world sequence mosh-server emits on every resize/reattach repaint (`CSI ?1003l ?1003h ?1004l ?1006l ?1006h`): tracking ends at `.anyEvent` and events flow SGR-encoded.
- **`decrstTrackingModeStillDisablesTracking`** — guard against overcorrecting: `?1003l` still turns tracking off even with an encoding active.

For docs to compare notes against: in xterm's ctlseqs the tracking modes (9, 1000–1003) and the coordinate extensions (1005, 1006, 1015, 1016) are documented as separate groups, and in the xterm source they're separate state variables — tracking drives `screen->send_mouse_pos` while the extensions set `screen->extend_coords` (button.c / charproc.c), so resetting an encoding can't touch tracking there. Easy empirical check in a real xterm: `printf '\e[?1003h\e[?1006h'`, then `printf '\e[?1006l'` — motion events keep arriving, just back in the default encoding. Alacritty/kitty/wezterm behave the same way.

migueldeicaza:
Thank you for all this information, I have checked it and it works are you described - I also appreciate your test updates!

PR #572, opened 2026-06-10
Add public searchMatchSummary for search-UI match counters

AlexDemzz:
## Summary

Adds a small public API on `TerminalView` to support match counters in search UIs:

```swift
public func searchMatchSummary (_ term: String, options: SearchOptions = SearchOptions(), limit: Int = 1000) -> (index: Int, total: Int)
```

- `index` is the 1-based position of the current match (the one selected by `findNext`/`findPrevious`) among all matches, or 0 when there is no current match.
- `total` is the overall match count, capped at `limit`.

This lets a client render the familiar "2/14" counter next to the find next/previous controls. `findNext`, `findPrevious`, and `clearSearch` are already public, but the underlying `SearchService.findAll`/`lastResult` are internal, so a counter can't be built from outside the package today.

## Implementation

A thin wrapper in `TerminalViewSearch.swift` over the existing `SearchService.findAll` port: it locates `search.lastResult` within the `findAll` results by (row, col). No changes to the search engine itself.

## Testing

- `swift build` clean on macOS.
- Exercised on iOS in a terminal app: counter tracks findNext/findPrevious correctly, including wrap-around at both buffer ends, and reports (0, 0)/(0, n) for empty/no-current-match states.

### DIFF
--- a/Sources/SwiftTerm/TerminalViewSearch.swift
+++ b/Sources/SwiftTerm/TerminalViewSearch.swift
@@ -51,6 +51,22 @@ extension TerminalView {
         return false
     }
 
+
+    /// Position of the current match among all matches for `term`: a 1-based
+    /// `index` (0 when there is no current match) and the `total` match count
+    /// (capped at `limit`). Drives a "2/14" style counter in a search UI.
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Lovely, thanks!

PR #574, opened 2026-06-10
LocalProcess: bound the PTY read backlog with backpressure

dxmone:
## Problem

`LocalProcess` re-arms its PTY read unconditionally, in two compounding ways:

1. **The read handler re-arms on every DispatchIO invocation, including partial deliveries** (`done == false`). DispatchIO invokes the handler several times per read op — partial deliveries followed by a final `done == true`. Since the handler issues a new `io.read` on each invocation, every partial delivery spawns an *additional concurrent read chain*. Under a fast producer the chains multiply, and hundreds of MB of in-flight reads pile up.

2. **No backpressure on the `pendingChunks` backlog** used by the main-queue delivery path. When the child writes faster than the consumer queue drains — easy whenever the main thread is busy with layout/rendering — the backlog grows without bound.

Measured with a `yes` flood against a busy main thread: **~280 MB/s of backlog growth**, multi-GB footprints in long sessions, eventually macOS's "out of application memory" termination.

## Fix

- Re-arm the next read only on `done == true`: one read op spawns exactly one successor.
- Track the backlog size in bytes. Past a **4 MB high-water mark**, stop re-arming the PTY read: the kernel PTY buffer fills and the child blocks in `write()` — exactly how every other terminal applies backpressure. `drainReceivedData` re-arms the read once the backlog drains below a **1 MB low-water mark**.

The synchronous (non-main-queue) delivery path already has natural backpressure via `dispatchQueue.sync` and is unchanged apart from the `done` gate.

## Measurements

Same `yes` flood, 10 s, headless terminal on the main queue:

| | before | after |
|---|---|---|
| footprint growth | unbounded (~280 MB/s) | plateaus at ~6 MB |
| behavior | backlog grows until OOM | continuous suspend/resume cycling, child blocks in kernel PTY buffer |

### DIFF
--- a/Sources/SwiftTerm/LocalProcess.swift
+++ b/Sources/SwiftTerm/LocalProcess.swift
@@ -93,6 +93,18 @@ public class LocalProcess {
     private var pendingChunkIndex: Int = 0
     private var pendingScheduled = false
     private let pendingLock = NSLock()
+    // Backpressure for the main-queue delivery path: without it the read loop
+    // re-arms unconditionally, so when the child produces output faster than
+    // the consumer queue drains it, pendingChunks grows without bound (observed
+    // ~280 MB/s with a `yes` flood against a busy main thread — multi-GB
+    // footprints in long sessions). Past the high-water mark we stop re-arming
+    // the PTY read; the kernel PTY buffer fills and the child blocks in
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Wow, very cute!

Issue #39, opened 2020-04-01
Mac: selection - when extending in word mode, it should extend by words


migueldeicaza:
Currently it only expands by letter, and should expand by word

migueldeicaza:
Implemented in 68062ef23955084ec0702661aa2b0cd6b861a4c2

PR #576, opened 2026-06-18
Mac: word-mode drag selection drops the seed word when dragging backwards

bones7456:
## Problem

On macOS, when a word is selected by **double-click** and the user then **drags backwards** (to the left, or up onto an earlier line), the double-clicked "seed" word is excluded from the selection.

### Steps to reproduce
With `aaa bbb ccc` on screen:

1. Double-click **`bbb`** → `bbb` is selected ✅
2. Drag left onto `aaa`

**Actual:** selection is `"aaa "` (seed word `bbb` dropped)
**Expected:** selection is `"aaa bbb"`

Dragging *forwards* (double-click `bbb`, drag onto `ccc`) already works and yields `"bbb ccc"`.

## Cause

Word-mode `dragExtend` only moves `end` and keeps `start` pinned at the seed word's start. On a backwards drag the resulting `start…end` range runs from the drag point up to the seed word's *start*, so the seed word is never included. This is in the word-extension selection added in #39.

## Fix

Record the double-clicked word range (`wordSelectionAnchor`) in `selectWordOrExpression`, and pivot the word-mode drag around it:

- drag goes **before** the word → extend `start` backward to a word boundary, keep the word's far `end`
- drag goes **after** the word → keep the word's `start`, extend `end` forward to a word boundary
- still **inside** the word → keep the whole word selected

Forward word-by-word extension is unchanged. The anchor is cleared whenever a non-word selection begins (`startSelection`, `select(row:)`, `selectNone`).

## Tests

Adds two regression tests to `SelectionTests`:

- `testWordDragExtendBackwardIncludesSeedWord` — drag left keeps the seed word (fails on `main`: gets `"aaa "`)
- `testWordDragExtendForwardExtendsByWords` — drag right still extends by words

All existing `SelectionTests` pass (`swift test --filter SelectionTests`, 23 tests).

### DIFF
--- a/Sources/SwiftTerm/SelectionService.swift
+++ b/Sources/SwiftTerm/SelectionService.swift
@@ -69,7 +69,16 @@ class SelectionService: CustomDebugStringConvertible {
     /**
      * Used to track the pivot point when selection in iOS-style selection
      */
-    public var pivot: Position? 
+    public var pivot: Position?
+
+    /**
+     * When `selectWordOrExpression` seeds a selection (a double-click), the
+     * selected range is recorded here so that a subsequent word-mode drag can
+     * pivot around the whole seed word.  Without this, dragging *backwards*
+     * past the seed word would leave `start` pinned at the seed word's start
+     * and the seed word would be dropped from the selection.
+     */
+    var wordSelectionAnchor: (start: Position, end: Position)?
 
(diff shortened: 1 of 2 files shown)


migueldeicaza:
Thank you!

PR #577, opened 2026-06-18
Fix terminal content hidden under the scroller after a font change

bones7456:
## Problem

After changing the font size (zoom in/out), the right-most columns of the
terminal are drawn **underneath the scroller** and become hidden. Zooming
the font in and back out also leaves the reported column count off by a
couple of columns from what a window resize produces.

## Root cause

The column/row count is computed in two places, and they disagree about
the scroller:

- The live window-resize path, `processSizeChange(newSize:)`, uses
  `getEffectiveWidth(size:)`, which subtracts the reserved scroller width.
- `resetFont()` (invoked when the font changes) computed the columns from
  the **raw** `frame.width`:

  ```swift
  let newCols = Int(frame.width / cellDimension.width)
  ```

So after a font change the terminal claims more columns than actually fit
in the visible (non-scroller) area, and the overflowing right-most columns
are rendered beneath the scroller.

## Fix

Compute the columns from `getEffectiveWidth(size:)` in `resetFont()` too,
matching `processSizeChange`. The function already exists and is used on
both macOS and iOS, so this keeps the two paths consistent and guarantees
content stays clear of the scroller.

## Tests

Added `FontResizeColumnsTests` (macOS) which:

- asserts the rendered width (`cols * cellDimension.width`) stays within
  `getEffectiveWidth(size:)` after a font change — i.e. content is not
  covered by the scroller;
- asserts the font-change path and the live-resize path report the same
  column count for the same frame.

Both tests fail on `main` (e.g. `renderedWidth 397.5 > usableWidth 383.0`,
and `53 != 51` columns) and pass with this change. The rest of the suite
is unaffected (the single pre-existing `testSynchronizedOutputBlocksDisplayUntilReset`
failure also reproduces on a clean `main` and is unrelated).

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -146,7 +146,11 @@ extension TerminalView {
         resetCaches()
         self.cellDimension = computeFontDimensions ()
         if (frame.width > 0) && (frame.height > 0) {
-            let newCols = Int(frame.width / cellDimension.width)
+            // Use getEffectiveWidth so the scroller's reserved width is taken
+            // into account, matching processSizeChange(). Computing columns from
+            // the raw frame width here would over-count by the scroller width,
+            // so zooming the font in and back out would drift the column count.
+            let newCols = Int(getEffectiveWidth(size: frame.size) / cellDimension.width)
             let newRows = Int(frame.height / cellDimension.height)
             resize(cols: newCols, rows: newRows)
         }
--- a/Tests/SwiftTermTests/FontResizeColumnsTests.swift
+++ b/Tests/SwiftTermTests/FontResizeColumnsTests.swift
@@ -1,0 +1,51 @@
+//
+//  FontResizeColumnsTests.swift
+//  SwiftTerm
+//
+//  Regression tests ensuring that after a font change the terminal content
+//  stays clear of the scroller, i.e. the recomputed geometry reserves the
+//  scroller width just like the live window-resize path does.
+//
+
+#if os(macOS)
+import Foundation
+import Testing
+import AppKit
+@testable import SwiftTerm
+
+final class FontResizeColumnsTests {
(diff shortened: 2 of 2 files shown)


bones7456:
Friendly ping on this one 🙂 Now rebased on latest `main` (sits right on top of #578, no conflicts). It's a small fix (+56/−1, 2 files) with macOS regression tests that fail on `main` and pass with the change. Its sibling PR #576 was already merged — this addresses the related scroller/column-drift issue after a font-size change. Happy to adjust anything if needed. Thanks for maintaining SwiftTerm!

bones7456:
The CI failure is the pre-existing flaky `testSynchronizedOutputBlocksDisplayUntilReset` (`SynchronizedOutputTests.swift:54`), unrelated to this change. The Mac/iOS/SPM builds all pass — only that one timing-dependent test records an issue.

The exact same failure shows up on `main` (e.g. run [27973775699](https://github.com/migueldeicaza/SwiftTerm/actions/runs/27973775699)): same test, same line :54, same assertion `(topLineText(...) → "NEW").hasPrefix("OLD")`. A re-run should go green.

bones7456:
Update on the CI failure: it's not flaky — `testSynchronizedOutputBlocksDisplayUntilReset` is a **stale test** that fails deterministically on `main` too. Commit 9446f60 removed the frozen `synchronizedOutputBuffer` snapshot (display blocking moved to the view layer) but never updated this test, so it still asserts the old snapshot behaviour.

Opened a separate test-only fix in #581 to keep it isolated from this font/scroller change. Once that merges, this PR's CI should go green on rebase.

bones7456:
Rebased onto latest `main` (now includes the #581 test fix) and CI is green ✅. The only thing previously blocking this was that stale test, so this should be ready for review/merge whenever you have a moment. Thanks!

migueldeicaza:
Apologies, let me review now

migueldeicaza:

PR #578, opened 2026-06-22
Center full-width (CJK) glyphs within their cell instead of left-pinning them

evertjr:
## Problem

Full-width (CJK) and other substituted glyphs are pinned to the left edge of their N-cell slot. The renderer discards CoreText's x advance and places every glyph at `column * cellDimension.width` on the cell grid, then draws it at natural size. For a single-width Latin glyph in a monospace font that's exact. But a full-width CJK glyph spans 2 cells, and with a typical Latin-metric base font the slot is wider than the glyph:

- SF Mono's advance is ≈ 0.62 em, so a 2-cell slot is ≈ 1.24 em.
- The substituted CJK glyph (Hiragino, PingFang, …) advances ≈ 1.0 em.

The ≈ 0.24 em of slack is dumped entirely on the **right** of every character, so each glyph sits hard against the left of its slot with a gap after it. It reads as if a space were inserted after every CJK character: Japanese/Chinese/Korean output looks "spaced out" and is markedly harder to read. Western text is unaffected.

This is structural, not letter-spacing: the gap is `N*cellWidth - glyphAdvance` for any Latin-metric base font. Terminal.app and Ghostty avoid the artifact by centering the glyph in its slot (Ghostty's glyph constraint, `align: .center`); the slack is identical but split symmetrically, which reads as natural CJK tracking. SwiftTerm pins left, which breaks the rhythm.

Reported by a Japanese user; reproduced with the default SF Mono on both the CoreGraphics and Metal renderers.

## Fix

Add `glyphSlotFit(font:glyph:columnWidth:)` on `TerminalView`, shared by the CoreGraphics (`drawTerminalContents`) and Metal glyph paths so they stay pixel-consistent. For a glyph in a slot of `columnWidth >= 2` it centers the glyph's advance box in the slot (`dx = (slotWidth - advance) / 2`); for the rare glyph whose ink overflows its slot it also scales it down to fit and re-centers it vertically. It returns identity for single-width cells, so it's a no-op for Latin and the hot path is untouched (the helper isn't even called for `columnWidth == 1`).

- Centering is by **advance, not ink**, so glyphs intentionally off-center in their em square (e.g. the CJK comma `、`) keep their designed position.
- Background fill, selection, and underline/strikethrough keep using the grid column; only the glyph ink moves within its own cell, so cursor/selection alignment is unchanged.
- The block caret is widened to span the full character and its glyph is centered the same way, so a block cursor covers the whole full-width character instead of just its left half, and the glyph doesn't jump when the caret lands on it. (Also fixes the pre-existing 1-cell caret on full-width characters.)

## Tests

Manually verified on both renderers:

- CJK (`日本語の表示、テスト。「カナ」 ABC 123`) now centers in its cells with no one-sided gap; mixed CJK+Latin stays grid-aligned; punctuation (`、。「」`) is balanced.
- Latin/ASCII is byte-for-byte unchanged (the helper short-circuits on single-width cells).
- Block cursor over CJK covers the whole character with no jump on blink/move; underline and selection stay cell-aligned.
- Cross-checked against a strict 1:2 font (Sarasa Mono): the offset is ≈ 0 (the glyph already fills the slot), confirming the helper only redistributes slack and stays out of the way when there's none.

Before:
SCR-20260621-sgrw

After:
SCR-20260621-sgta

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -83,6 +83,31 @@ struct ViewLineInfo {
     var boxDrawings: [BoxDrawingRenderItem]
 }
 
+/// How to place a single glyph within its `columnWidth`-cell slot.
+///
+/// Full-width (CJK) and other substituted glyphs would otherwise be pinned to
+/// the left edge of their slot, which dumps the slack between the glyph's
+/// advance and the slot width onto the right of every cell — the phantom gap
+/// reported for CJK text. ``GlyphSlotFit`` centers the glyph's advance box in
+/// the slot so the slack is split evenly instead, matching Terminal.app and the
+/// centering Ghostty applies to wide glyphs (its glyph constraint with
+/// `align: .center`). As a safety net it also constrains a glyph whose ink
+/// overflows its slot — too wide for the columns, or too tall for the cell —
+/// scaling it down to fit and re-centering it vertically. That fit-to-cell guard
+/// is in the spirit of Ghostty's `size` constraint; it is not Ghostty's
+/// cap-height CJK down-scaling (#8709), which corrects an up-scaling we never do.
+/// `dx`/`dy` are point offsets applied to the glyph origin; `scale` is a uniform
+/// downscale (`<= 1`).
+struct GlyphSlotFit {
+    var dx: CGFloat = 0
+    var dy: CGFloat = 0
+    var scale: CGFloat = 1
+
+    static let identity = GlyphSlotFit()
+
+    var isIdentity: Bool { dx == 0 && dy == 0 && scale == 1 }
+}
+
 extension TerminalView {
     typealias CellDimension = CGSize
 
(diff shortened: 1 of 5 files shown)


migueldeicaza:
Wow, very nice find - thank you for the report and the patch.

I am reviewing.

migueldeicaza:
I only have one concern, the caret code does not seem to be in sync, I have not tested this, but probably we should test it.   

CoreGraphics caret path applies fit.dx/fit.dy but ignores fit.scale. Normal text rendering scales oversized glyphs down via CTFontCreateCopyWithAttributes, and the Metal cursor path scales the quad, but the AppKit/UIKit CaretView still draws the original-size glyph. For any glyph that triggers the new overflow safety net, the block cursor rendering maybclip or visibly jump relative to the surrounding text. 

Shouldn't we mirror the normal CoreGraphics path and draw with a scaled CTFont when fit.scale is less than one?

evertjr:
Great observation, this was indeed a missing piece. I managed to create an evaluation test and the cursor glyph didn't match the text

Just added a new commit addressing this.

migueldeicaza:
Thank you!

PR #579, opened 2026-06-22
macOS: don't yank the viewport to the bottom while scrolled up

gilbert-barajas:
## Problem

When new output arrives while the user has scrolled up into the scrollback, the viewport jumps back to the bottom — yanking them away from what they were reading. Follow-tail should only auto-scroll when the viewport is already pinned to the bottom.

## Fix

Follow the tail (auto-scroll on new output) only when the user is already at the bottom; otherwise preserve the current scroll position.

## Tests

`TerminalFollowTailTests` covers both the scrolled-up case (position preserved) and the at-bottom case (follows new output).

## Note

Split out of #555 (live-resize reflow) for review hygiene — it was an unrelated change riding along. Includes a one-line `SyncDebug` no-op so the package builds (it's referenced on `main` but never committed); that build-unblocker is shared with #555 and will drop out on rebase once either lands.

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -1925,6 +1925,16 @@ extension TerminalView {
     public func scrollTo (row: Int, notifyAccessibility: Bool = true)
     {
         let displayBuffer = terminal.displayBuffer
+        // Drive the feed-follow gate from the real scroll position: parked above
+        // the bottom ⇒ userScrolling, so streaming output no longer yanks the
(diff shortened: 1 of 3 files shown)


migueldeicaza:
Mhm, we have had a few attempts at addressing this, let me see if I can find my notes on the other attempts - although I think they were mostly iOS based.

migueldeicaza:
I took a look at this, since I had a previous effort in a branch, but it was not quite complete and needed a few tuneups.   I am happy now with what we have in iOS and I think it works fine on MacOS, can you take it for a spin?   I think this is now fixed.

If it is not, please reopen the bug.

PR #582, opened 2026-06-24
Clear the dirty region in the CoreGraphics renderer (fixes restricted-region scroll artifacts)

tavvet:
### Problem
Scrolling or editing inside a restricted DECSTBM scroll region (a fixed header and/or footer, as in `nano`, `vim`, `less`, `htop`) leaves stale rows and a sub-cell "ghost" on the region's last row on the **CoreGraphics renderer** (Metal disabled). The Metal path masks it by clearing the whole target each frame.

Reproducible: open `nano` on a file with lines wider than the window and hold PageDown/PageUp; or do `vim` line edits in a fixed-footer layout.

### Root cause — the renderer, not the scroll primitives
The scroll/edit primitives report a correct dirty range. The CG renderer:

1. **never clears the dirty region's background** before painting — `drawTerminalContents` fills background per-cell-run only, and **only for cells that carry an explicit background**; default-background cells rely on transparent backing-store pixels showing the layer's `backgroundColor`. AppKit clears the backing store only on a full-view redraw, so a partial repaint (a restricted region, IL/DL) keeps stale glyphs/backgrounds.
2. **only invalidates the sub-cell bottom remainder when `rowEnd == rows-1`** — a restricted region ends at `scrollBottom < rows-1`, so the sliver below it is never repainted → the ghost.

(There is no pixel blit; unchanged pixels persist via AppKit's dirty-rect coalescing over the layer backing store.)

### Fix — macOS CG path only; Metal unaffected
In `AppleTerminalView`:

1. `context.clear(dirtyRect)` before the per-row paint — clears the invalidated region to transparent (a clear, **not** a fill, so a translucent background is preserved), giving the per-cell paint a clean slate.
2. Extend a mid-screen region's invalidated `region` down by one cell, so the bottom remainder is cleared too (previously only `rowEnd == rows-1` got this).

This fixes **every** row-mutating primitive (`scroll`, `reverseIndex`, `cmdScrollUp/Down`, `cmdInsertLines/DeleteLines`) at the source. With the renderer clearing any dirtied region, the primitives only need to flag the rows they changed — so `refreshScrolledRegion` flags just `[scrollTop, scrollBottom]` (keeping the full-screen + scrollback cheap path) instead of inflating to the whole viewport.

The Metal renderer is untouched: it already `loadAction = .clear`s the whole target each frame, and the `region` extension only feeds the non-Metal `setNeedsDisplay`.

### Testing
Verified in `nano` / `vim` / `less` / `htop` on the CG renderer — no stale rows or bottom ghost while paging or editing — with a translucent background intact, and normal full-screen output scrolling unaffected (cheap path preserved). The package builds cleanly.

---
*This supersedes an earlier per-primitive workaround on this branch (which inflated each scroll op's dirty range to the whole viewport) with the renderer-level root fix.*

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -1284,6 +1284,16 @@ extension TerminalView {
         }
         var placeholderImageCache: [UInt32: TTImage] = [:]
 
+        #if os(macOS)
+        // Clear the invalidated region before painting. We fill only cells that carry
+        // an explicit background; default-background cells rely on transparent backing-
+        // store pixels showing the layer's background color. AppKit clears the backing
+        // store only on a full-view redraw, so a partial repaint (a restricted DECSTBM
+        // scroll region, line insert/delete) otherwise keeps stale glyphs/backgrounds.
+        // Clear to transparent — not fill — so a translucent background is preserved.
+        context.clear(dirtyRect)
+        #endif
+
         for row in firstRow...lastRow {
             if row < 0 {
                 continue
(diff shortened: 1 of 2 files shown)


migueldeicaza:
Thank you, let me review this - it looks pretty clean.

migueldeicaza:
This looks good!

PR #589, opened 2026-07-12
Bidirectional text (Arabic/Hebrew) rendering: UAX #9 reordering, Arabic shaping, terminal-wg escapes

naifalziadna:
## Summary

This adds bidirectional text support (Arabic, Hebrew) to SwiftTerm, following the [terminal-wg BiDi recommendation](https://terminal-wg.pages.freedesktop.org/bidi/)'s implicit model (the approach used by mlterm and VTE): the terminal buffer stays in strict logical order, and rows containing RTL content get a visual permutation at render time. BiDi-unaware apps (ls, cat, shells) display Arabic/Hebrew correctly; BiDi-aware apps can opt out via the standard escapes.

**Now rebased onto current main** (single commit, conflict-free) and reduced to BiDi-only — the unrelated fixes and downstream configuration that were previously mixed in are gone (fixes live in #590; the scrollback bits were my app's configuration and did not belong here).

## Try it

```sh
cat Tests/bidi-sample.txt
```

The sample covers plain Arabic/Hebrew, mixed-direction lines with digits, lam-alef ligatures, vocalized text (harakat/niqqud), bracket mirroring, and a long soft-wrapping Arabic paragraph. For a reference rendering, open the same file in a browser (browsers implement full UAX #9); for "before", cat it on a build of main — everything renders left-to-right in logical order with disconnected Arabic letters.

### BiDi engine (`Apple/TerminalBidi.swift`, new)
- UAX #9 reordering computed per row via a CoreText probe line (one UTF-16 unit per cell), producing logical↔visual column maps.
- Arabic contextual shaping into presentation forms, lam-alef ligatures, harakat/diacritic preservation.
- Joining context carried across soft-wrapped rows (`BufferLine.isWrapped`) so words shaped across a wrap seam stay joined.
- Bracket mirroring (UAX #9 L4); box-drawing mirroring behind DECSET 2500.
- Paragraph direction: per-row autodetect (first strong character), or forced LTR/RTL/off.

### Renderer integration
- `buildAttributedString` walks rows in visual order with per-cell display substitutions; a utf16→cell map plus `CTRunGetStringIndices` keeps every glyph (combining marks, ligature clusters) anchored to its column.
- Arabic-script cells render as isolated column-anchored segments so font-level required ligatures (e.g. الله) cannot collapse columns.
- Caret drawing and mouse hit-testing translate through the visual maps.

### terminal-wg escape sequences
- **BDSM** (`CSI 8 h/l`): implicit vs explicit BiDi support mode.
- **SPD** (`CSI Ps SP S`): presentation direction, `0` = LTR, `3` = RTL (plain `CSI Ps S` scroll-up unaffected).
- **DECSET/DECRST 2501**: paragraph-direction autodetection; **2500**: box-drawing mirroring.
- All report through DECRQM; RIS restores defaults.
- Public API: `bidiParagraphDirection` (embedder override) and `effectiveBidiDirection` (override resolved against the escape state).

## Testing

`swift test` passes: 469 tests, including new suites for the engine (reordering, shaping, wrap-seam joining, mirroring), the render path, and the escape sequences.

## Notes for review

- Render-time only; the emulation layer is untouched except for the new mode state on `Terminal`.
- The spec leaves the autodetection default to the implementation; this PR defaults it **on** so RTL works out of the box (VTE ships it off — happy to flip for parity).
- Mode changes re-render the visible buffer with the current state; VTE stamps direction per paragraph as written. Per-paragraph stamping would need buffer-side state — proposed as follow-up work if this lands.

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -66,9 +66,21 @@ struct ViewLineSegment {
     let columnWidth: Int
     let characterCount: Int
     let attributedString: NSAttributedString
-    
+    /// Maps every UTF-16 unit of attributedString to the ordinal of the cell
+    /// it belongs to, so glyphs can be positioned by cell (combining marks
+    /// share their base's cell instead of shifting the column grid).
+    let utf16ToCellOrdinal: [Int]
+
     var columnSpan: Int {
         return max(0, characterCount * columnWidth)
+    }
+
+    @inline(__always)
+    func cellOrdinal(forUTF16 index: Int) -> Int {
+        if index >= 0 && index < utf16ToCellOrdinal.count {
+            return utf16ToCellOrdinal[index]
+        }
+        return max(0, utf16ToCellOrdinal.last ?? 0)
     }
 }
 
@@ -109,6 +121,26 @@ struct GlyphSlotFit {
 }
 
 extension TerminalView {
+
+    /// The paragraph direction actually used for rendering: the
+    /// user-selected `bidiParagraphDirection` unless it is `.auto`, in which
+    /// case the escape-driven terminal-wg state applies — BDSM explicit mode
+    /// (`CSI 8 l`) turns BiDi off, autodetection (`DECSET 2501`) selects
+    /// per-row detection, and otherwise the SPD direction (`CSI Ps SP S`)
+    /// forces LTR or RTL paragraphs.
+    public var effectiveBidiDirection: BidiParagraphDirection {
+        if bidiParagraphDirection != .auto {
+            return bidiParagraphDirection
+        }
+        if !terminal.bidiSupportEnabled {
+            return .off
+        }
+        if terminal.bidiAutodetectDirection {
+            return .auto
+        }
+        return terminal.bidiRTLPreference ? .rightToLeft : .leftToRight
+    }
+
     typealias CellDimension = CGSize
 
 #if os(macOS)
@@ -658,6 +690,8 @@ extension TerminalView {
         let columnWidth: Int
         private var attributedString = NSMutableAttributedString()
         private var characterCount: Int = 0
+        private var utf16ToCellOrdinal: [Int] = []
+        private var cellCount: Int = 0
         
         init(column: Int, columnWidth: Int) {
             self.column = column
@@ -668,16 +702,25 @@ extension TerminalView {
             characterCount == 0
         }
         
-        mutating func append(text: String, attributes: [NSAttributedString.Key: Any]) {
+        /// Appends a batch of text; `cellUTF16Lengths` holds one entry per
+        /// terminal cell in the batch (its text length in UTF-16 units).
+        mutating func append(text: String, attributes: [NSAttributedString.Key: Any],
+                             cellUTF16Lengths: [Int]) {
             attributedString.append(NSAttributedString(string: text, attributes: attributes))
             characterCount += 1
+            for length in cellUTF16Lengths {
+                for _ in 0..<max(1, length) {
+                    utf16ToCellOrdinal.append(cellCount)
+                }
+                cellCount += 1
+            }
         }
         
         func buildIfNeeded() -> ViewLineSegment? {
             guard !isEmpty else {
                 return nil
             }
-            return ViewLineSegment(column: column, columnWidth: columnWidth, characterCount: characterCount, attributedString: attributedString)
+            return ViewLineSegment(column: column, columnWidth: columnWidth, characterCount: characterCount, attributedString: attributedString, utf16ToCellOrdinal: utf16ToCellOrdinal)
         }
     }
     
@@ -690,6 +733,33 @@ extension TerminalView {
         let selectionColumns = selectedColumnsRange(row: row, cols: cols)
         var col = 0
         var builder: ViewLineSegmentBuilder?
+
+        // BiDi: rows containing RTL content are walked in visual order with
+        // per-cell display substitutions (shaped Arabic forms, mirrored
+        // brackets); the layout is nil for plain LTR rows, which use the
+        // logical path below unchanged. Soft-wrapped lines pass their edge
+        // characters as joining context so shaping survives the wrap seam.
+        var precedingScalar: UInt32? = nil
+        var followingScalar: UInt32? = nil
+        let bidiDirection = effectiveBidiDirection
+        if bidiDirection != .off {
+            let lines = terminal.displayBuffer.lines
+            if line.isWrapped, row > 0, row - 1 < lines.count {
+                precedingScalar = TerminalBidi.trailingScalar(of: lines[row - 1], cols: cols,
+                                                              terminal: terminal)
+            }
+            if row + 1 < lines.count, lines[row + 1].isWrapped {
+                followingScalar = TerminalBidi.leadingScalar(of: lines[row + 1], cols: cols,
+                                                             terminal: terminal)
+            }
+        }
+        let bidiLayout = TerminalBidi.layout(line: line, cols: cols, terminal: terminal,
+                                             direction: bidiDirection, font: fontSet.normal,
+                                             precedingScalar: precedingScalar,
+                                             followingScalar: followingScalar)
+        let bidiWritingDirectionKey = NSAttributedString.Key(kCTWritingDirectionAttributeName as String)
+        var visualCol = 0
+        var visualIndex = 0
         var kittyPlaceholders: [KittyPlaceholderCell] = []
         var previousPlaceholder: KittyPlaceholderCell?
         var previousPlaceholderAttribute: Attribute?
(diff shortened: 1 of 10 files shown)


migueldeicaza:
Wow, this is fabulous, and I am very supportive of this effort.   Thank you for taking the time to explore the space and submit this contribution.   

Could you share some sample text that I could "cat" into a terminal so I can compare before and after, and so I can compare how things look like in a browser?

Some quick notes, and I will start to familiarize myself with the changes in the meantime:

1. Procedurally, the code currently contains conflicts, please update so we can review
2. The code contains changes to scroll back history, is there a reason to have this here?
3. Would love to get your bug fixes that are part of this PR separated, as those seems like good fixes to get right away while this review continues.

naifalziadna:
Thank you for the kind words and the quick look! All three points addressed:

**Sample text** — I added a cat-able showcase to the branch: `cat Tests/bidi-sample.txt`. It covers plain Arabic/Hebrew, mixed-direction lines with digits, lam-alef, vocalized text (harakat/niqqud), bracket mirroring, and a long soft-wrapping Arabic paragraph. For comparison: open the same file in a browser (full UAX #9 reference), and cat it on a build of `main` for the "before" (logical order, disconnected letters). I avoided pasting the text inline here because the browser bidi-reorders it in the comment, which defeats the before/after purpose.

**Conflicts** — resolved. I rebased onto current main and squashed to a single conflict-free commit. Nice surprise while doing it: several of the fixes I had bundled (SGR release/motion flag, OSC 52 clipboard delivery, `setTerminalFocus`, find-panel actions) have since landed on main independently, so they simply dropped out.

**Scrollback changes** — good catch; those were my downstream app's configuration (a `defaultScrollbackLines` static and a scroll-wheel variant) that leaked into the branch. Removed — this PR is now BiDi-only.

**Separated fixes** — opened as #590: the DECSET 1004 initial focus report, the mouse *motion* events reporting buffer-absolute rows (press/release/drag already convert to viewport rows), and a stray `print` in the SGR-pixel encoder. Those are the only ones from the original bundle not already on main.

migueldeicaza:
Wow, the selection now works really well!    Still reviewing, just testing now the code.  

It looks like our rendering was readable, but the selection and positioning were very wrong.

migueldeicaza:
Update: I am familiarizing myself with the upstream specification.

migueldeicaza:
From the top of my head, I think that it would be worth enabling this by default, and get folks to opt out of it, so we can identify any performance regressions.   My only reservation is that from reading the spec, that we would need to do paragraph-level processing (which this patch only currently does by treating each line as a paragraph), so maybe that more precise mode needs to be opt-in (or if the performance is negligible, we can keep on).

I am taking the spec at face value that it is good, and that other smarter folks have reviewed it and agree it is good - as this is very new territory for me.  So my feedback is between the patch and the spec.

Some of my feedback:

1. We do not really use UTF-16 internally, we use Grapheme clusters (What Swift calls a 'Characters', and the surfacing of UTF16 in the patch just gave the air that something might be off.  Would love to understand the rationale behind that, and whether we should revisit that in those terms.

2. Paragraph handling: The current implementation resolves UAX #9 independently for every BufferLine. Soft-wrapped rows are not independent paragraphs: they form one paragraph until a hard newline. This can cause a continuation row to select a different base direction from its first row, and weak/neutral characters around the wrap boundary can resolve incorrectly. Passing adjacent scalars fixes Arabic joining only; it does not provide the paragraph context required by UAX #9.

SwiftTerm already has the paragraph-boundary information needed here: a row with isWrapped == true continues the previous row, and a paragraph ends when the next row is absent or has isWrapped == false. The PR currently uses this only for adjacent Arabic joining context. Could it use the same backward/forward scan already used by implicit link detection to collect and lay out the complete paragraph?

2. The `effectiveBiDi` seems to be collapsing the six different options from the terminal-wg spec into fewer options, I think that we would need this to be include implicit/explicit, auto detection and direction, no?   Otherwise we would lose information (the .auto value means that we still need to delegate work to the caller, it would not know if the fallback is left to right or right to left.   The .off value also seems to be able to just represent explicit LTR, but no way to state explicit RTL).

4. BiDi state is global and retroactively applied, rather than being captured per paragraph. This prevents paragraphs created under different modes from coexisting and allows mode changes intended for future output to alter existing soft-wrapped paragraphs and scrollback. 

7. The box rendering is tricky.   DECSET 2500 computes a mirrored box-drawing override, but the default custom box renderer draws ch.code, which is the original unmirrored character. The override works only when custom box rendering is disabled. Could BoxDrawingRenderItem.codePoint be derived from the resolved display character when one exists? A focused render/build test should verify that an RTL-resolved ┌ produces a render item for ┐ while box mirroring is enabled.

8. The spec points out different escape sequences for settings the modes than the one this patch implements for setting LTR/RTL which are CSI x SPACE k, and we are using CSI x SPACE S versions.   Not sure why.

Some of these I think I can fix, but I could use your expertise as you know more about this than me.

migueldeicaza:
In addition to my comments above, I asked Claude to code review it, and found some interesting nuggets:

  1. Soft reset (DECSTR) doesn't restore BiDi state — Sources/SwiftTerm/Terminal.swift:3525
  resetToInitialState now resets bidiSupportEnabled/bidiAutodetectDirection/bidiRTLPreference/bidiBoxMirroring, but cmdSoftReset was not updated to
  match. A full-screen app that sends CSI 8 l (BDSM explicit mode, disabling reordering) and then issues DECSTR on exit leaves
  bidiSupportEnabled=false. All subsequent RTL shell output renders scrambled (logical, unshaped order) with no way for the shell to recover short
  of a hard reset.

  2. iOS tap/selection doesn't map visual→logical columns — Sources/SwiftTerm/iOS/iOSTerminalView.swift:682 (also :677)
  The parallel fix that MacTerminalView.getPositionFromView received in this PR was not applied to iOS. calculateTapHit treats the tapped visual
  column as the logical column. On any RTL row, tapping/dragging selects, highlights, and copies the wrong characters, and the insertion
  point/magnifier lands on the wrong cell. Affects selection (631/741/762/817) and UITextInput hit-testing (1123/1149/1400).

  3. VoiceOver output is garbled for RTL lines — Sources/SwiftTerm/iOS/iOSTerminalView.swift:2991
  accessibilityAttributedLine reuses the BiDi-reordered buildAttributedString, so the emitted NSAttributedString is in visual order rather than
  logical order. VoiceOver/Braille reads Arabic/Hebrew mis-ordered (RTL reversed, brackets mirrored, lam-alef blanked) instead of in logical
  reading order.

  4. defaultScrollbackLines is dead API with a misleading doc comment — Sources/SwiftTerm/iOS/iOSTerminalView.swift:1287
  New public static, documented as "read during setup and whenever the font changes," but never read anywhere. setupOptions builds
  TerminalOptions(cols:rows:) which defaults scrollback to 500. A user setting defaultScrollbackLines = 5000 silently keeps only 500 lines —
  contradicting the documented behavior.

  Performance regression (single root cause)

  5. BiDi layout runs on every row, every frame — even for pure-LTR terminals — Sources/SwiftTerm/Apple/AppleTerminalView.swift:756 (also :105,
  :2031, MacTerminalView.swift:2412)
  effectiveBidiDirection returns .auto by default (both flags default true). buildAttributedString unconditionally calls TerminalBidi.layout, which
  allocates a cells array and calls terminal.getCharacter(for:) for every cell just to scan for RTL triggers before hitting the nil fast path. On
  large screens at high refresh this adds a per-row, per-frame allocation + character-decode pass to the universal CoreText render path —
  CPU/battery cost for all users, including English-only. Worth a cheap early-out (e.g. a per-row "contains any codepoint ≥ RTL range" scalar
  check) before the full extract.

  BiDi shaping correctness (cleanup-tagged)

  6. Lam-alef ligature drops combining marks — Sources/SwiftTerm/Apple/TerminalBidi.swift:445
  When lam (U+0644) precedes an alef variant, the alef cell is overwritten with result[i + 1] = " ". Any combining harakat on that alef cell are
  silently dropped, so vocalized lam-alef renders without its diacritic.

  7. Repair pass indexes possibly-RTL-unset neighbors — Sources/SwiftTerm/Apple/TerminalBidi.swift:555
  For a cell whose probe produced no glyph (control/formatting char), the repair pass reinserts using rtl[k-1]/rtl[k+1]. Those default to false
  when the neighbor was itself a no-glyph cell, so a formatting char between two RTL cells can land on the wrong side of its neighbor.

  Minor cleanup

  8. Redundant intermediate-byte validation — Sources/SwiftTerm/Terminal.swift:3282
  cmdSelectPresentationDirection's guard collect == [32] else { return } is unreachable — the parser only routes 0x53 here when collect == [32].
  Duplicated dispatch condition that must be kept in sync across two sites.

  ---
  Refuted / not real (verifier rejected 4): the Mac hit-test "maps to column 0" pair, the caret-remap-to-0 claim, and a redundant-else in
  AppleTerminalView.swift:847.

  The two highest-impact items are #1 (soft reset) and #2 (iOS selection) — both are straightforward to fix by mirroring what the PR already did
  elsewhere (hard reset for #1, the Mac hit-test fix for #2). Want me to apply any of these?

migueldeicaza:
I have a PR work in progress with the updates changes: https://github.com/migueldeicaza/SwiftTerm/pull/621

migueldeicaza:
Ok landed the alternative PR based on this one.  It contains the various issues addressed, and a test suite and tools to validate visually.

it also matches the performance of main without regressing and I was inspired so it is now also faster thatbwhat we had

PR #590, opened 2026-07-12
Mouse motion row and focus reporting fixes

naifalziadna:
Split out from #589 as requested — the bug fixes from that branch that are not already on main (the SGR release/motion-flag fix, OSC 52 clipboard delivery, `setTerminalFocus`, and `performFindPanelAction` all landed independently since I branched).

**1. DECSET 1004 reports nothing on enable.** xterm reports the current focus state as soon as an application enables focus reporting; without it, applications that enable 1004 while the terminal is already focused believe they are unfocused until the first real focus change. Concretely: Claude Code gates its Enter handling on focus, so pressing Enter did nothing until you clicked away and back. `Terminal` now tracks the host view's focus (`setTerminalFocus` records it even while reporting is off) and `DECSET 1004` replies with `CSI I`/`CSI O` right away. Covered by new tests (enable focused, enable unfocused, silent when disabled).

**2. Mouse motion events report buffer-absolute rows.** `sendMotion` in the Mac view received `hit.grid.row`, which includes `yDisp`, while the press/release/drag paths already convert to screen rows. Once any scrollback exists, hover-tracking applications receive rows far outside the viewport and their hit-testing never matches (observed as dead hover targets in TUIs with clickable rows).

**3. Stray debug `print` in the SGR-pixel mouse encoder** fires on every event; removed.

Full suite passes (444 tests).

### DIFF
--- a/Sources/SwiftTerm/Mac/MacTerminalView.swift
+++ b/Sources/SwiftTerm/Mac/MacTerminalView.swift
@@ -2682,7 +2682,14 @@ open class TerminalView: NSView, NSTextInputClient, NSUserInterfaceValidations, 
         
         if terminal.mouseMode.sendMotionEvent() {
             let flags = encodeMouseEvent(with: event, overwriteRelease: true)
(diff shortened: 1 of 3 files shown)


migueldeicaza:
Thank you for sending this patch along!

PR #593, opened 2026-07-14
Fix Metal renderer crash in packaged apps (Bundle.module fatalErrors)

### LINKED ISSUE #592: Metal renderer crashes (fatalError) in any packaged .app that isn't the build machine
### Summary
Enabling the Metal renderer (`setUseMetal(true)`) crashes the whole process with `fatalError` on any machine other than the one that built the app, if the app is a packaged `.app` bundle (not `swift run`).

### Root cause
`MetalTerminalRenderer.candidateBundles()` (`Sources/SwiftTerm/Apple/Metal/MetalTerminalRenderer.swift`) puts `Bundle.module` first in its list of places to look for `SwiftTerm_SwiftTerm.bundle` (the shader source/metallib):

```swift
private static func candidateBundles() -> [Bundle] {
    var bundles: [Bundle] = []
    #if SWIFT_PACKAGE
    bundles.append(Bundle.module)
    #endif
    bundles.append(Bundle(for: MetalTerminalRenderer.self))
    bundles.append(Bundle.main)
    return bundles
}
```

The problem is that `Bundle.module` isn't a normal, safe API here — for an *executable* target, SwiftPM generates an accessor that looks like this:

```swift
extension Foundation.Bundle {
    static nonisolated let module: Bundle = {
        let mainPath = Bundle.main.bundleURL.appendingPathComponent("SwiftTerm_SwiftTerm.bundle").path
        let buildPath = "/absolute/path/baked/in/at/compile/time/.build/.../SwiftTerm_SwiftTerm.bundle"
        let preferredBundle = Bundle(path: mainPath)
        guard let bundle = preferredBundle ?? Bundle(path: buildPath) else {
            Swift.fatalError("could not load resource bundle: from \(mainPath) or \(buildPath)")
        }
        return bundle
    }()
}
```

Two things go wrong for any real packaged `.app`:
1. `mainPath` checks `Bundle.main.bundleURL` — the `.app`'s *root* — not `Contents/Resources`, which is where a packaged app's build script actually has to put an SPM resource bundle for it to be inside the signed bundle at all.
2. The fallback is the *build machine's own absolute `.build` path*, baked into the binary at compile time. This only "works" on the machine that built the release, which is exactly backwards for a distributed binary.

Since `fatalError` aborts the process rather than throwing or returning nil, merely *evaluating* `Bundle.module` is what crashes — `candidateBundles()`'s other two entries (`Bundle(for:)`, `Bundle.main`) never get a chance to run as intended fallbacks.

### Repro
1. Build any macOS app that depends on this package via SPM, ships `SwiftTerm_SwiftTerm.bundle` correctly in `Contents/Resources` (the only valid location in a signed `.app`), and calls `terminalView.setUseMetal(true)`.
2. Run the built `.app` on a machine other than the one that built it (or just delete the build machine's `.build` directory to simulate this).
3. The process crashes immediately with `EXC_BREAKPOINT (SIGTRAP)` / `fatalError("could not load resource bundle: ...")`, the instant Metal is enabled.

Found this while shipping [Portside](https://github.com/mcglothi/portside), a macOS terminal app — had to disable the Metal renderer toggle entirely in packaged builds to avoid crashing every user except the developer.

### Suggested fix
Opening a PR now that replaces the `Bundle.module` candidate with a manual, nil-safe probe using `Bundle(url:)` (a normal failable initializer — verified it returns nil for a missing path rather than crashing) at both `Bundle.main.bundleURL` (matching the generated accessor, for bare `swift run` binaries) and `Bundle.main.resourceURL` (the location a packaged `.app` actually uses, which the generated accessor never checks).

mcglothi:
Fixes #592.

## Problem

`MetalTerminalRenderer.candidateBundles()` puts `Bundle.module` first when looking for `SwiftTerm_SwiftTerm.bundle`. For an executable target, SwiftPM's generated `Bundle.module` accessor `fatalError`s (aborts the process, doesn't throw/return nil) if it can't find the bundle at either of its two hardcoded candidates:

1. `Bundle.main.bundleURL` — the `.app`'s root, not `Contents/Resources` where a packaged app's build actually has to place an SPM resource bundle to ship it inside a signed `.app` at all.
2. The build machine's absolute `.build/.../SwiftTerm_SwiftTerm.bundle` path, baked into the binary at compile time.

So any app that enables the Metal renderer crashes on every machine except the one that built it — and since the crash happens just from *evaluating* `Bundle.module`, `candidateBundles()`'s other fallback candidates (`Bundle(for:)`, `Bundle.main`) never run.

## Fix

Replace the `Bundle.module` candidate with a manual probe using `Bundle(url:)` — a normal failable initializer, verified to return `nil` for a missing path rather than crashing — checked at:
- `Bundle.main.bundleURL` (matches the generated accessor's own lookup, for bare `swift run` binaries)
- `Bundle.main.resourceURL` (the `Contents/Resources` location a packaged `.app` actually uses, which the generated accessor never checks)

Both are additive probes before the existing `Bundle(for:)`/`Bundle.main` fallbacks, so behavior is unchanged for any environment where the old code didn't crash, and strictly better (an actual chance to find the bundle in a real packaged app) where it did.

## Testing

- `swift build` passes on this branch.
- Verified in isolation that `Bundle(url:)` returns `nil` for a nonexistent path and a working bundle (with resource lookup working) for a real one — the core assumption the fix relies on, unlike `Bundle.module`'s generated accessor.
- Reproduced and worked around the original crash downstream in [Portside](https://github.com/mcglothi/portside) (currently disables the Metal toggle entirely in packaged builds pending this); happy to verify this fix directly against that app's packaging pipeline if useful.

### DIFF
--- a/Sources/SwiftTerm/Apple/Metal/MetalTerminalRenderer.swift
+++ b/Sources/SwiftTerm/Apple/Metal/MetalTerminalRenderer.swift
@@ -2770,7 +2770,30 @@ final class MetalTerminalRenderer: NSObject, MTKViewDelegate {
     private static func candidateBundles() -> [Bundle] {
         var bundles: [Bundle] = []
         #if SWIFT_PACKAGE
-        bundles.append(Bundle.module)
+        // Deliberately not `Bundle.module`: SwiftPM's generated accessor for
+        // executable targets calls `fatalError` (aborting the whole process,
+        // not throwing) instead of returning nil when it can't locate
+        // `SwiftTerm_SwiftTerm.bundle`. Its only two candidates are
+        // `Bundle.main.bundleURL` (the .app's *root*, not `Contents/Resources`
+        // where a packaged .app actually puts SwiftPM resource bundles) and
+        // the build machine's absolute `.build/.../SwiftTerm_SwiftTerm.bundle`
+        // path baked into the binary at compile time. That means any app that
+        // bundles this package and ships the resource bundle correctly still
+        // crashes the instant Metal rendering is requested on a machine other
+        // than the one that built it. Probe for the same bundle name
+        // ourselves — mirroring the accessor's own main-bundle-relative
+        // lookup, plus the `Contents/Resources` location it misses — so a
+        // missing bundle falls through to the next candidate instead of
+        // aborting the process.
+        let bundleName = "SwiftTerm_SwiftTerm.bundle"
+        if let url = Bundle.main.resourceURL?.appendingPathComponent(bundleName),
+           let resourceBundle = Bundle(url: url) {
+            bundles.append(resourceBundle)
+        }
+        if let url = Bundle.main.bundleURL.appendingPathComponent(bundleName) as URL?,
+           let resourceBundle = Bundle(url: url) {
+            bundles.append(resourceBundle)
+        }
         #endif
         bundles.append(Bundle(for: MetalTerminalRenderer.self))
         bundles.append(Bundle.main)


migueldeicaza:
Thank you!

mcglothi:
Verified against Portside's actual packaging pipeline (`Contents/Resources`) — with `SwiftTerm_SwiftTerm.bundle` copied there alongside our own resource bundle, the Metal renderer now initializes correctly in the packaged `.app` with no crash. Thanks for the quick merge! Just shipped it in [Portside 0.7.3](https://github.com/mcglothi/portside/releases/tag/v0.7.3).

PR #594, opened 2026-07-15
Detect quoted file paths with spaces as implicit links

bones7456:
## Problem

Implicit link detection uses the Ghostty-style regex, which only continues past a space when the next segment ends in `/` or `.`. A path like:

```
'/Users/me/Screenshot 2026-07-15 at 09.58.24.png'
```

is therefore only matched up to `Screenshot`, even though the surrounding quotes make the boundary unambiguous — this is exactly how shells and many CLI tools print paths containing spaces.

## Change

Add a pre-pass in `implicitLinkMatch` over the already-built implicit line map: when the lookup target sits inside a `'...'` or `"..."` pair whose content starts with `/`, `~/`, `./` or `../`, the whole quoted content (quotes excluded) is returned as the link.

Details:
- **Opener anchoring instead of strict pairing** — any quote directly followed by a path-looking prefix is treated as an opener, closed by the nearest quote of the same kind. This keeps detection robust when the surrounding prose contains apostrophes (`don't miss '/tmp/a b.png'`).
- **Innermost candidate wins** — nested quoting like `"'/a b.png'"` resolves to `/a b.png`.
- The cell-offset → row-ranges mapping is factored out of `implicitLinkMatch` into a shared `implicitMatch(in:text:startOffset:endOffset:)`, used by both the regex path and the new quoted path, so hover highlighting and click resolve the same full range (including across wrapped lines).
- Non-path quoted strings (`'hello world'`) and clicks outside the quoted range still return no match; regex behavior is unchanged otherwise.

## Tests

8 new tests in `LinkLookupTests` covering single/double quotes, `~/` paths, nested quotes, wrapped lines, apostrophes in prose, and negative cases. Full suite passes (452 tests).

### DIFF
--- a/Sources/SwiftTerm/Terminal.swift
+++ b/Sources/SwiftTerm/Terminal.swift
@@ -7314,26 +7314,39 @@ open class Terminal {
         )
     }
 
-    /// Paths that contain spaces defeat the Ghostty-style regex, but
-    /// shell-quoted output gives an unambiguous boundary: when the lookup
-    /// target sits inside a '...' or "..." pair whose content looks like a
-    /// filesystem path, the whole quoted content (quotes excluded) is the
-    /// link. Innermost wins so "'/a b.png'" resolves to /a b.png.
+    /// Quote pairs that delimit a path in program output, as the opener mapped
+    /// to the closers that can end it. Shells and most tools use the straight
+    /// pairs; GNU tools quote as `like this'; markdown-flavored output (and the
+    /// AI agents that emit it) uses `like this`; anything that has been through
+    /// typographic substitution uses the curly pairs.
+    private static let pathQuoteClosers: [Character: Set<Character>] = [
+        "'": ["'"],
+        "\"": ["\""],
+        "`": ["`", "'"],
+        "\u{2018}": ["\u{2019}"],
+        "\u{201C}": ["\u{201D}"]
+    ]
+
+    /// Paths that contain spaces defeat the Ghostty-style regex, but quoted
+    /// output gives an unambiguous boundary: when the lookup target sits
+    /// inside a quote pair whose content looks like a filesystem path, the
+    /// whole quoted content (quotes excluded) is the link. Innermost wins so
+    /// "'/a b.png'" resolves to /a b.png.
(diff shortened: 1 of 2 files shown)


bones7456:
Pushed a follow-up commit that also improves the *unquoted* spaces case: the Ghostty-style space-segment rule required each space-joined segment to **end** with `/` or `.`, so greedy backtracking matched `/x/face cropped.png` only up to `cropped.`, silently dropping the extension. The segment pattern now allows trailing word characters after the last `/`/`.` (`[\/.]` → `[\/.]\w*`), so the full `face cropped.png` is matched. Prose words after a path (containing neither `/` nor `.`) are still excluded, and existing behavior is otherwise unchanged — full suite passes.

migueldeicaza:
I need to think about this, the idea of the link is that it represents URLs, not arbitrary file paths, and while certainly some apps use '/path' others might use "/path" or `path' or the typographical versions of those.

bones7456:
Rebased on main and handled the quoting conventions you listed — `` `path' `` (GNU) and the curly pairs, plus `` `path` `` since markdown-flavored output is where most agent tools put paths. The "closed by the same character" rule is now an opener → closers map, so an opener pairs only with its own closer and a mismatched pair isn't a boundary.

On links being URLs rather than arbitrary file paths: the implicit branch already detects bare paths (that's the Ghostty port), and `1483b09` now opens them, `file.swift:13` included. So the open question isn't whether a path can be a link, it's where one *ends* — and a quote pair is the only unambiguous answer available. Everything else stays heuristic.

Two notes on the tests, since it's easy to write ones that prove nothing here:

- They use `/tmp/dir.d with prose.txt` and click inside `with`. An unquoted space-joined segment only counts when it carries a `/` or `.` of its own, so that click finds nothing without a quote pair — `testUnquotedPathStopsAtProseSegment` pins that. A path like `/tmp/my file.txt` passes all of them unquoted, which is why I switched away from it.
- The dotless branch (`anyPathSpaceSegments`, pre-existing) joins any space-separated run, so `/tmp/dir with prose.txt ok` currently swallows the trailing ` ok`. Not from this PR and I left it alone; happy to look at it separately. Quoted boundaries do fix that case.

Full suite (699 tests) passes.

PR #595, opened 2026-07-15
Mac: open implicit file-path links instead of failing with paramErr (-50)

bones7456:
## Problem

Cmd-clicking an implicitly detected filesystem path (e.g. `/Users/me/file.png`) pops the Finder dialog **"The application can't be opened. -50"**.

The default `requestOpenLink` implementation builds every link with `URL(string:)`. Implicit detection returns raw paths, so this produces a *scheme-less* URL, which Launch Services rejects with `paramErr` (OSStatus -50). Paths containing spaces fail `URL(string:)` outright, so clicking them silently does nothing.

## Change

- Add `TerminalView.openDefaultLink(_:)`: links with a URL scheme open as before; scheme-less links are treated as filesystem paths and opened via `URL(fileURLWithPath:)`, with tilde expansion and an existence check (so nonexistent/relative paths become a no-op instead of an error dialog).
- The `TerminalViewDelegate` protocol-extension default now delegates to that helper.
- `LocalProcessTerminalView` gets a real `open func requestOpenLink` member. The protocol-extension default cannot be overridden from app code (the conformance witness is resolved statically), so embedding apps previously had no way to customize link handling — e.g. to resolve relative paths against the shell's actual cwd. Now subclasses can.

Complements #594 (quoted-path detection): the paths with spaces that #594 starts detecting can only be opened with this fix, though each PR stands on its own.

### DIFF
--- a/Sources/SwiftTerm/Mac/MacLocalTerminalView.swift
+++ b/Sources/SwiftTerm/Mac/MacLocalTerminalView.swift
@@ -156,6 +156,15 @@ open class LocalProcessTerminalView: TerminalView, TerminalViewDelegate, LocalPr
     open func rangeChanged(source: TerminalView, startY: Int, endY: Int) {
         //
     }
+
+    /// Implementation of the TerminalViewDelegate method. Being a class
+    /// member (unlike the protocol-extension default) it can be overridden
+    /// by subclasses, for example to resolve relative paths against the
+    /// shell's current working directory.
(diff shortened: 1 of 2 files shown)


migueldeicaza:
This conflicted with another patch, but also, did not handle things like "demo.swift:13", these have been fixed.

migueldeicaza:
I landed the alternative patch.

PR #599, opened 2026-07-18
Let subclasses customize link and key handling

moshegottlieb:
Two customization points that are currently closed off on open classes.

`requestOpenLink` — `LocalProcessTerminalView` sets `terminalDelegate = self`, so link activation resolves to the `TerminalViewDelegate` protocol extension default. Protocol extension methods are statically dispatched, so a subclass can't override link handling at all — embedders have to swap the delegate out at click time to work around it. This adds an open implementation on the class and extracts the existing open logic into an internal openLink helper called from both, so there's no duplication.

`pressesBegan` (iOS) — `TerminalView` is an open class, but `pressesBegan` is declared public override, preventing subclasses outside the module from customizing hardware keyboard handling (e.g. app-level ⌘-shortcuts over a focused terminal). UIResponder declares it open; this just restores the inherited access level.

No API removals and no behavior change for existing users.


migueldeicaza:
Makes sense, thanks for sending this along!

PR #600, opened 2026-07-18
Line-accurate scroll wheel with optional sensitivity control (macOS)

klc:
## Problem

On macOS, fast trackpad flicks make the terminal **skip** whole blocks of lines: the viewport jumps ahead and you have to scroll slowly to see the content it flew past. Slow scrolling is fine; only fast/momentum scrolling exhibits it.

## Root cause

`scrollWheel(with:)` feeds `abs(event.deltaY)` through `calcScrollingVelocity`, a step function:

```swift
if delta > 9 { return max(terminal.rows, 20) }  // a whole page
if delta > 5 { return 10 }
if delta > 1 { return 3 }
return 1
```

During a fast flick, `deltaY` spikes above 9, so a *single* event scrolls `max(rows, 20)` lines at once — a full page jump. The intermediate lines are never rendered, which is exactly the "skipping" users see. It also ignores the pixel-precise deltas macOS provides for trackpads, so scrolling never feels 1:1.

## Fix

Replace the step function with pixel-delta accumulation:

- **Precise (trackpad) deltas** — accumulate `scrollingDeltaY` and convert to whole lines by cell height, carrying the sub-cell remainder to the next event. This gives smooth, true 1:1 scrolling with no page jumps.
- **Classic wheel deltas** — map `scrollingDeltaY` to line units directly, guaranteeing a notch always moves at least one line.

All three existing branches (mouse reporting, alternate buffer, normal buffer) are preserved and driven by the same computed line count; scroll direction semantics are unchanged.

## New: `scrollSensitivity`

Adds a public multiplier:

```swift
public var scrollSensitivity: CGFloat = 1.0   // clamped to a 0.05 floor
```

`1.0` (the default) is the system-native rate, so **behavior is unchanged for anyone who doesn't set it**. Host apps can expose it as a user preference (e.g. a "scroll speed" slider) — line-based terminal scrolling often feels faster than native momentum, so being able to dial it down is useful.

## Compatibility & scope

- Single file, macOS only (`Sources/SwiftTerm/Mac/MacTerminalView.swift`).
- No public API removed; `scrollSensitivity` is purely additive with a backward-compatible default.
- The drag-selection auto-scroll velocity (which also used `calcScrollingVelocity`) is intentionally left as-is — that function is retained and unchanged, only moved.

## Testing

Built and exercised on macOS with both a trackpad and a mouse: fast flicks now scroll smoothly line-by-line instead of jumping pages, mouse-wheel notches move predictably, and alternate-buffer apps (vim/less) still receive arrow-key scrolling. Lowering `scrollSensitivity` slows scrolling proportionally without ever stalling.

### DIFF
--- a/Sources/SwiftTerm/Mac/MacTerminalView.swift
+++ b/Sources/SwiftTerm/Mac/MacTerminalView.swift
@@ -2693,40 +2693,8 @@ open class TerminalView: NSView, NSTextInputClient, NSUserInterfaceValidations, 
         }
     }
     
-    public override func scrollWheel(with event: NSEvent) {
-        if event.deltaY == 0 {
-            return
-        }
-        if allowMouseReporting && !shiftBypassesMouseReporting(for: event) && terminal.mouseMode != .off {
-            let hit = calculateMouseHit(with: event)
-            let displayBuffer = terminal.displayBuffer
-            let screenRow = max (0, min (displayBuffer.rows - 1, hit.grid.row - displayBuffer.yDisp))
-            let button = event.deltaY > 0 ? 4 : 5
-            let flags = event.modifierFlags
-            let buttonFlags = terminal.encodeButton(button: button, release: false, shift: flags.contains(.shift), meta: flags.contains(.option), control: flags.contains(.control))
-            let lines = calcScrollingVelocity(delta: Int(abs(event.deltaY)))
-            for _ in 0..<lines {
-                terminal.sendEvent(buttonFlags: buttonFlags, x: hit.grid.col, y: screenRow, pixelX: hit.pixels.col, pixelY: hit.pixels.row)
-            }
-        } else if terminal.isDisplayBufferAlternate {
-            let lines = calcScrollingVelocity(delta: Int(abs(event.deltaY)))
-            for _ in 0..<lines {
-                if event.deltaY > 0 {
-                    sendKeyUp()
-                } else {
-                    sendKeyDown()
-                }
-            }
-        } else {
-            let velocity = calcScrollingVelocity(delta: Int(abs(event.deltaY)))
-            if event.deltaY > 0 {
-                scrollUp(lines: velocity)
-            } else {
-                scrollDown(lines: velocity)
-            }
-        }
-    }
-    
+    /// Velocity curve used by drag-selection auto-scroll (how fast to scroll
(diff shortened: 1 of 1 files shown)


Sources/SwiftTerm/Mac/MacTerminalView.swift
```diff
@@ -2693,54 +2693,92 @@ open class TerminalView: NSView, NSTextInputClient, NSUserInterfaceValidations,
...
+    /// system's native rate; values below `1.0` slow scrolling down, above speed
+    /// it up. Host apps can expose this as a user preference. Clamped to a small
+    /// positive floor so it can never disable scrolling entirely.
+    public var scrollSensitivity: CGFloat = 1.0 {
+        didSet { scrollSensitivity = max(0.05, scrollSensitivity) }
+    }
+
     public override func scrollWheel(with event: NSEvent) {
-        if event.deltaY == 0 {
+        guard let cellHeight = cellDimension?.height, cellHeight > 0 else {
```
migueldeicaza, line 2731: This patch lost the case for deltaY == 0, which I guess is fine, but we should use the event.scrollingDeltaY != 0 or we risk turning a 0 into a -1 below.

migueldeicaza (changes requested):

Sources/SwiftTerm/Mac/MacTerminalView.swift
klc: Good catch — fixed in a47b356.

You're right that the rewrite silently dropped the guard. The precise/trackpad path happened to be safe (a zero delta adds nothing to the accumulator, `lines` comes out 0, and we return), but the non-precise branch is exactly the hazard you describe: `rounded` is 0, and the "a wheel notch must always move at least one line" fallback turns that into -1, scrolling on an event that carried no movement.

Restored it as an early exit against `scrollingDeltaY`, the delta this method now reads:

```swift
public override func scrollWheel(with event: NSEvent) {
    if event.scrollingDeltaY == 0 {
        return
    }
    guard let cellHeight = cellDimension?.height, cellHeight > 0 else {
        return
    }
```

Same semantics as the original `if event.deltaY == 0 { return }`, just keyed to the delta we consume. Builds clean.

migueldeicaza:
Thank you so much!

PR #601, opened 2026-07-19
Fixed flipped mouse reporting toggle

moshegottlieb:
Mouse reporting toggle color is flipped on initialization.
It starts as blue which **later** spells out as _allows mouse reporting **OFF**_ but mouse reporting starts as **ON**.

| Tap # | `allowMouseReporting` | Highlight (before) | Highlight (after) |
|-------|-----------------------|--------------------|-------------------|
| 0     | on                    | blue ❌            | off ✅            |
| 1     | off                   | blue ✅            | blue ✅           |
| 2     | on                    | off ✅             | off ✅            |


migueldeicaza:
Thank you for the patch!

PR #603, opened 2026-07-20
Mac: repaint the rows where screen updates actually land while scrolled back

bones7456:
Rebased on top of #620, which fixed the CoreGraphics half of this while the PR was open by invalidating the whole view when `yDisp != yBase`.

What this changes:

- **CG**: invalidate the rows the change actually lands on instead of the whole view. Rows below the viewport invalidate nothing; a full-screen refresh still covers everything visible.
- **Metal**: the dirty range used `yDisp + row` where the absolute row is `yBase + row`. Now derived from the same view rows as the CG path, so the full-screen rule matches too — it was marking `31...43` dirty for a visible `29...41`.
- **Bidi** (new since I opened this): `renderingDependencyRange` is asked about `yDisp + rowStart`, the same conflation a third time. Now `yBase`-relative, unioned into the view rows so both paths inherit it.
- **`metalDirtyRange` was replaced, not merged**, so a link-highlight row queued before the renderer drew was dropped. Highlights don't bump the line generation, so the renderer's cache check can't recover it.
- **`invalidateLinkHighlightRow`** no longer routes through `updateRange`, which can't express a scrollback row at all.

8 tests, 4 per path. The Metal ones attach a bare `MTKView(device: nil)` instead of calling `setUseMetal(true)` — the range bookkeeping never draws, so no GPU is needed and they actually run on CI. Gating on a real device makes them silently skip under `swift test`: the shader bundle sits next to the `.xctest` bundle, which `candidateBundles()` doesn't probe, so `setUseMetal(true)` throws `shaderSourceMissing`.

#620's test passes unchanged. Full suite (690 tests) passes, iOS builds.

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -1166,12 +1166,33 @@ extension TerminalView {
 
     func invalidateLinkHighlightRow(_ bufferRow: Int)
     {
+        // The highlighted row can live in the scrollback (visible while the
+        // user has scrolled up), which the update-range machinery cannot
+        // express — its rows are relative to the live screen (yBase). Since
+        // the highlight is view state (no buffer line changes), invalidate
+        // the displayed row directly instead.
         let displayBuffer = terminal.displayBuffer
-        let screenRow = bufferRow - displayBuffer.yDisp
-        guard screenRow >= 0 && screenRow < terminal.rows else {
+        let viewRow = bufferRow - displayBuffer.yDisp
+        guard viewRow >= 0 && viewRow < terminal.rows else {
             return
         }
-        terminal.updateRange(borrowing: displayBuffer, screenRow)
+        #if canImport(MetalKit)
+        if metalView != nil {
+            if let current = metalDirtyRange {
+                metalDirtyRange = min (current.lowerBound, bufferRow)...max (current.upperBound, bufferRow)
+            } else {
+                metalDirtyRange = bufferRow...bufferRow
+            }
+            requestMetalDisplay()
+            return
+        }
+        #endif
+        #if os(macOS)
+        let rowY = frame.height - (CGFloat(viewRow) + 1) * cellDimension.height
+        setNeedsDisplay(CGRect (x: 0, y: rowY, width: frame.width, height: cellDimension.height))
+        #else
+        setNeedsDisplay(bounds)
+        #endif
     }
 
     func linkVisibleForClick(match: Terminal.LinkMatch, hasCommandModifier: Bool) -> Bool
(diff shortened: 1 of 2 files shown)


migueldeicaza:
I had codex review the changes, just as a safety check:

[P2] Preserve pending Metal dirty rows. [updateDisplay](https://github.com/migueldeicaza/SwiftTerm/blob/43358308ad6b5062745f4afceb45ce242e7c6e93/Sources/SwiftTerm/Apple/AppleTerminalView.swift#L1850-L1858) replaces metalDirtyRange. This can discard a link-highlight row queued at [lines 923–930](https://github.com/migueldeicaza/SwiftTerm/blob/43358308ad6b5062745f4afceb45ce242e7c6e93/Sources/SwiftTerm/Apple/AppleTerminalView.swift#L923-L930) before Metal draws it. Link highlights do not change the buffer-line generation, so the renderer does not recover the lost update. Merge the terminal range with the existing dirty range.

[P2] Apply the full-screen rule to Metal. The CoreGraphics path treats refresh(0, rows - 1) as a visible-viewport refresh. The Metal path instead translates that range from yBase. When scrolled back two rows, my test got 31...43 instead of the visible 29...41. Use visibleStart...visibleEnd for the same full-screen condition and add a Metal version of the regression test.

bones7456:
Thanks — both hold up, and the second one is a regression I introduced. Fixed in 5134c24 (added on top rather than squashed, so the delta is reviewable).

**Preserving pending Metal dirty rows.** Confirmed the mechanism: `requestMetalDisplay()` only marks the MTKView as needing display, the renderer consumes and clears the range at draw time, and a link highlight doesn't touch the buffer line's generation — so `cacheValid` stays true and the lost row is unrecoverable. Worth noting this was self-inflicted too: before this PR `invalidateLinkHighlightRow` went through `terminal.updateRange`, which merged; writing `metalDirtyRange` directly is what made replacement lossy. There is now an `addMetalDirtyRows` helper used by the highlight path, `updateDisplay`, and the iOS branch (which replaced too — it writes the whole visible range, so nothing was actually lost there, but the semantics are the same now).

I left `MacTerminalView.selectionChanged` alone: it also replaces, but with the entire visible range, and a queued highlight row is visible by construction, so it's a superset.

**Applying the full-screen rule to Metal.** Your numbers reproduce exactly. Also worth flagging that this case was *right* before this PR by accident: `yDisp + rowStart` happens to equal the visible range for `refresh(0, rows-1)`, and `updateFullScreen()` sets `refreshEnd = rows`, which tripped the old `rowEnd < terminal.rows` guard and fell through to the visible range. My switch to `yBase` broke the first of those.

Rather than special-casing the full-screen condition twice, the Metal branch now derives its absolute rows from the view rows the CoreGraphics path already computes (`yDisp + viewStart ... yDisp + viewEnd`). That gets all three cases from one place — partial update lands at `yBase + row`, full-screen refresh equals the visible range, rows pushed below the viewport add nothing — and it removed the nested out-of-range fallbacks, so the branch is net smaller.

**Tests.** Four Metal cases; three of them fail against 4335830 (the full-screen one with precisely your `31...43` vs `29...41`).

One thing worth knowing if you write Metal tests later: my first attempt used `try view.setUseMetal(true)` and skipped when Metal was unavailable — and it passed against the *buggy* code, because it was silently skipping. Under `swift test` the device is fine but shader loading isn't: `SwiftTerm_SwiftTerm.bundle` sits in `.build/debug/` as a *sibling* of the `.xctest` bundle, and `candidateBundles()` probes `Bundle.main`'s two locations and `Bundle(for:)` itself, but not the directory containing the code bundle — so `setUseMetal(true)` throws `shaderSourceMissing` and any GPU-gated test is quietly green. The tests now attach a bare `MTKView(frame:device: nil)` instead: the dirty-range bookkeeping never draws, so no device, renderer or shader library is involved and the tests actually run on CI.

That `candidateBundles()` gap looks like a real one beyond tests — an app that ships the resource bundle next to a framework rather than in `Contents/Resources` would hit it too. Happy to send a separate PR adding the code-bundle-sibling candidate if you want it; it felt unrelated to this one.

Full suite (460 tests) passes, and the package still builds for iOS.

bones7456:
Rebased. The conflict wasn't textual — #620 landed the same diagnosis meanwhile — so to be explicit about what this still changes:

#620 invalidates the whole view when scrolled back. This narrows it to the rows that changed, which is the case its own message calls out: a TUI redrawing a status block while you read scrollback repaints one band instead of the entire view on every update. `inPlaceRepaintWhileScrolledBackInvalidatesRenderedRow` passes unchanged — it asserts the rendered row is repainted, which the precise rect satisfies.

The rest is what #620 doesn't reach: the Metal dirty range and its full-screen rule, the link-highlight invalidation, and the bidi dependency lookup added since, which asks `renderingDependencyRange` about `yDisp + rowStart`. That's the same conflation a third time — masked on the CG side by the whole-view invalidation, but `absoluteDependencyRange` went straight to the Metal renderer.

If you'd rather leave #620's behavior alone, I'm happy to drop the CG narrowing and keep only the Metal, highlight and bidi fixes.

PR #605, opened 2026-07-22
Render Powerline separators as cell-aligned geometry

futantan:
## What changed

- custom-render `U+E0B0`, `U+E0B2`, `U+E0B4`, and `U+E0B6` when `customBlockGlyphs` is enabled
- share normalized separator geometry across CoreGraphics and Metal
- preserve resolved foreground colors for selection, inverse, and dim states
- render Powerline glyphs correctly under block cursors and DEC double-width/double-height line modes
- add one-device-pixel join overdraw to prevent seams without vertical cell overflow

## Why

Powerline separators rendered through Core Text can cover one physical pixel outside the terminal cell while the cell background remains pixel-aligned. This produces visible steps at the top or bottom of Starship and Powerline segments, and font-outline changes do not reliably solve it across fonts, sizes, scales, and rendering backends.

Treating these cell-joining symbols as terminal graphics primitives makes their bounds deterministic and follows the same approach already used for box drawing and block elements.

## Validation

- `swift test --filter PowerlineRendererTests`: 5 tests, 0 failures
- `swift test`: 59 XCTest tests and 452 Swift Testing tests, 0 failures
- verified CoreGraphics and Metal code paths
- verified iOS/macOS block cursor paths
- verified 2x/3x geometry and DEC double-width join behavior
- real iPhone 17 Pro Simulator at 3x over Docker SSH, fish, Starship 1.26.0, and the official Gruvbox Rainbow preset
- before: separator solid-color bounds were 61 physical pixels high while the segment cell was 60 pixels
- after: all six sampled segment colors were exactly 60 physical pixels high

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -81,6 +81,7 @@ struct ViewLineInfo {
     var kittyPlaceholders: [KittyPlaceholderCell]
     var blockElements: [BlockElementRenderItem]
     var boxDrawings: [BoxDrawingRenderItem]
+    var powerlineGlyphs: [PowerlineRenderItem]
 }
 
 /// How to place a single glyph within its `columnWidth`-cell slot.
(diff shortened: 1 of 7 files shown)


migueldeicaza:
Very nice, and it fits in the existing architecture - I think the only thing that is slightly different is that the existing blocks do honor the no-antialias flag, and this one does not seem to.

Let me review in depth.

migueldeicaza:
Ok, this looks really good, I won't commit tonight, but will do tomorrow or Monday morning.

migueldeicaza:
Ok reviewed, I have a minor nit, but I can fix after (the fact that the property that drives the name for enable/disable this rendering talks about blocks, but it is now a little bit more than that, and the antialiasing being always on here).

PR #607, opened 2026-07-22
decode Kitty placeholder IDs in RGB order

sirfergy:
Kitty Unicode placeholders encode truecolor image and placement IDs in RGB byte order. Decoding them in reverse stores the transmitted PNG under one ID while placeholder cells look up another, leaving correctly-sized blank regions.

## What
- Decode truecolor placeholder IDs as `(red << 16) | (green << 8) | blue`.
- Cover asymmetric image/placement IDs, high-byte composition, and transmitted-image lookup.

## Tests
The full SwiftTerm suite passes (436 tests), including a golden case captured from Copilot CLI: APC ID `0x42AC8A` with placeholder RGB `42 AC 8A`.


migueldeicaza:
Thank you for tracking this down and sending a PR!

PR #608, opened 2026-07-24
Share one VT500 transition table across parsers; clamp oversized CSI/DCS/OSC params

klc:
Two related hardening changes to `EscapeSequenceParser`.

### 1. Reuse a single VT500 transition table across parser instances
The VT500 transition table is immutable once built, but a fresh copy was constructed for every parser instance. Build it once as a static and share it — cheaper parser creation, less allocation.

### 2. Clamp oversized CSI/DCS/OSC numeric parameters instead of resetting to zero
Numeric parameter parsing computed `pars[last] * 10 + digit` and, once the running value crossed an overflow guard, silently reset the whole parameter to `0`. Because the accumulator was reset before it could grow large enough to overflow the multiply, this never actually trapped — but it did turn legitimately huge values (e.g. a crafted sequence like `\x1b[99999999999999999999999m`) into `0` instead of "very large".

The new `appendingParameterDigit` checks *before* multiplying and clamps to `UInt16.max` (the largest value the parser ever needs) on overflow — matching xterm's behavior of capping oversized parameters. Applied to the CSI param shortcut, the DCS/APC accumulator, and `parseInt` (OSC).

### Testing
Added `CsiParameterParsingTests`. Full suite green (456 tests).

### DIFF
--- a/Sources/SwiftTerm/EscapeSequenceParser.swift
+++ b/Sources/SwiftTerm/EscapeSequenceParser.swift
@@ -87,7 +87,7 @@ enum ParserAction : UInt8 {
 class TransitionTable {
     // data is packed like this:
     // currentState << 8 | characterCode  -->  action << 4 | nextState
-    var table: [UInt8]
+    private(set) var table: [UInt8]
     
     init (len: Int)
     {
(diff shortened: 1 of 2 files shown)


migueldeicaza:
Oh, interesting, I had not considered that.   I'll check what xterm's behavior is.

migueldeicaza:
I took a slightly different path based on your patch, but with some changes:

- CSI/DCS values will cap at UInt16.max 
- CSI/DCS sequences exceeding 24 parameters are dropped (like Ghostty does)
- Overflow state persists correctly across input chunks.
- OSC selectors now reject malformed values and true Int overflow instead of aliasing to OSC 65535.
- OSC 4 palette indices use the strict decimal parser instead.
- Made transition table immutable after construction (so we get a let instead of a var)

Thank you!

migueldeicaza:
The patch that landed: https://github.com/migueldeicaza/SwiftTerm/commit/d327a2bb8b74e2d9cd79838e755b8931d1bf4975

PR #609, opened 2026-07-24
CircularList: replace abort()/force-unwraps with precondition checks

klc:
`CircularList` used `abort()` and force-unwraps (`!`) on out-of-range access. `abort()` terminates the process with no diagnostic, and the force-unwraps trap with a generic message.

Replace them with `precondition` checks that carry a clear message and use standard trap semantics. No behavior change on the valid path — only the failure mode for programmer error is clearer.

### Testing
Full suite green (452 tests).


migueldeicaza:
Thank you for this cleanup!

PR #610, opened 2026-07-24
Avoid force-casting CTRun font/color attributes

klc:
Reading font and foreground-color attributes out of a `CTRun`'s attribute dictionary used force-casts (`as!`), which crash if an attribute is missing or has an unexpected type.

Use conditional casts (`as?`) with sensible fallbacks instead, so an unexpected attribute dictionary degrades gracefully rather than crashing. Touches `AppleTerminalView` and `CaretView`.

### Testing
Full suite green (452 tests).


migueldeicaza:
Thanks for the contribution, I would love to know under which scenarios this happens.

PR #611, opened 2026-07-24
Fix cross-terminal hyperlink GC and make TinyAtom thread-safe

klc:
Three related fixes around hyperlink payload atoms (`TinyAtom`) in `CharData`.

### 1. Make `TinyAtom` thread-safe
The global atom table (`intern` / `removeValue`) was mutated without synchronization. Concurrent access from multiple terminals could corrupt the dictionary and crash (observed as a `SIGABRT` inside `removeValue`). Guard the table with a lock.

### 2. Track hyperlink payload codes per `Terminal`
Hyperlink-atom garbage collection tracked live codes globally, so tearing down one `Terminal` could collect atoms still referenced by another live `Terminal`. Track referenced codes per `Terminal` so GC only reclaims atoms that terminal actually owns.

### 3. Remove dead `TinyAtom.lastCollected`
No longer read after change (2); removed.

### Testing
Added coverage in `LinkLookupTests`. Full suite green (454 tests).

### DIFF
--- a/Sources/SwiftTerm/CharData.swift
+++ b/Sources/SwiftTerm/CharData.swift
@@ -192,7 +192,7 @@ public struct TinyAtom {
     var code: UInt16
     private static let lock = NSLock()
     private static var map: [UInt16:Any] = [:]
-    static var lastUsed: UInt16 = 0
+    private static var lastUsed: UInt16 = 0
     static var lastCollected: Int = 0
     static let empty = TinyAtom (code: 0)
    
@@ -217,9 +217,16 @@ public struct TinyAtom {
     }
     
     public static func release(code: UInt16) {
+        release(codes: [code])
+    }
+
+    static func release<S: Sequence>(codes: S) where S.Element == UInt16 {
         lock.lock()
         defer { lock.unlock() }
-        map.removeValue(forKey: code)
+
+        for code in codes where code != 0 {
+            map.removeValue(forKey: code)
+        }
     }
     
     /// Returns the target for the TinyAtom
--- a/Sources/SwiftTerm/CharData.swift
+++ b/Sources/SwiftTerm/CharData.swift
@@ -190,8 +190,9 @@ public struct Attribute: Equatable, Hashable {
 /// it could in theory be changed to be 24 bits without much trouble
 public struct TinyAtom {
     var code: UInt16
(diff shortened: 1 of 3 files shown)


migueldeicaza:
Oops, thanks for catching this - a year or two ago, I did an audit for these kinds of things when doing stress testing, but I guess the stress testing never covered this.

Reviewing.

migueldeicaza:
Ok, this looks good, but one small challenge: Publicly attached payloads are no longer garbage-collected. 

Let me think how to solve this.

*Edit* clarification: the issue is that the API is public, so atoms can be created by consumers of the library, not just the library.

klc:
Thanks! One clarification that might make this easier to decide: the old sweep didn't really collect publicly-attached payloads either — it freed them out from under the caller.

```swift
for code in UInt16(TinyAtom.lastCollected + 1)...UInt16(TinyAtom.lastUsed) {
    if used.contains(code) { break }
    TinyAtom.release(code: code)
}
```

`used` is only the *current* terminal's buffers, so an atom a consumer created via the public `TinyAtom.lookup` but hadn't attached to a cell yet was eligible for release on the next `garbageCollectPayload()` of any terminal — sometimes before its first use. So this PR removes an accidental premature free rather than a working feature.

The question that decides the fix: does a single atom need to be referenced by more than one live `Terminal`, or held by a consumer outside any buffer?

- If no: `TinyAtom.release(code:)` is already public, so caller-owned lifetime is effectively the existing contract — it just isn't documented. I can add a `Terminal`-level intern-and-register entry point (interns the value and records it in `payloadCodes`) so consumer payloads attached to a terminal get collected by that terminal, and document `TinyAtom.lookup` as caller-owned. Small diff.
- If yes: refcounting the atom table is the sound answer — `Terminal` retains a code when it adopts it and releases on GC/deinit, and the entry is freed at zero. Bigger change, your call.

(Adopting unknown codes seen during the buffer scan doesn't work on its own: two terminals showing the same code would both adopt it, and the first to drop it frees a payload the other still displays — the exact cross-terminal case this PR fixes. It only works on top of refcounts.)

Happy to implement either in this PR.

migueldeicaza (approved):
Ok while this breaks the public api, it was a very rare use.   So I think we can survive by adding a deprecated notice for users to fix.

PR #612, opened 2026-07-26
Let SwiftTerm cross-compile from macOS to Linux

ASNNetworks:
Two independent blockers hit when building with the [Swift Static Linux SDK](https://www.swift.org/documentation/articles/static-linux-getting-started.html) from macOS:

```
swift build --swift-sdk x86_64-swift-linux-musl
```

Neither changes behaviour on macOS, iOS, native Linux or Windows.

### 1. The Apple/Mac/iOS excludes don't apply when cross-compiling

`Package.swift` excludes those directories behind `#if os(Linux) || os(Windows)`. But a package manifest is compiled and **run on the host**, so when cross-compiling from a Mac that is false, `platformExcludes` stays empty, and the Apple sources are handed to the Linux target:

```
Sources/SwiftTerm/Apple/CaretView.swift:9:8: error: no such module 'CoreText'
```

A manifest cannot see the destination platform, so this adds an explicit opt-in rather than trying to detect it:

```
SWIFTTERM_EXCLUDE_APPLE=1 swift build --swift-sdk x86_64-swift-linux-musl
```

Native Linux and Windows builds are unaffected — they still take the `#if` branch.

### 2. `KittyGraphics.swift` assumes glibc on Linux

It imports `Glibc` under `os(Linux)`, but the Static Linux SDK builds against musl, where the C library module is `Musl` and `Glibc` does not exist:

```
Sources/SwiftTerm/KittyGraphics.swift:9:8: error: no such module 'Glibc'
```

Switched to `canImport` so the cascade covers musl, glibc and Windows without altering any existing case.

### Verified

Built `x86_64-swift-linux-musl` from macOS (Swift 6.2.4) and ran the resulting statically linked binary on Ubuntu 24.04 with no Swift installed. macOS builds are unchanged.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -3,10 +3,17 @@
 import PackageDescription
 import Foundation
 
+// A package manifest is compiled and run on the HOST, so `os(Linux)` is false
+// when cross-compiling from macOS to Linux — and the Apple/Mac/iOS sources are
+// then handed to the Linux target, which fails on `import CoreText`. There is
+// no way for a manifest to see the destination, so allow the exclude to be
+// forced explicitly.
+let excludeAppleSources =
+    ProcessInfo.processInfo.environment["SWIFTTERM_EXCLUDE_APPLE"] == "1"
 #if os(Linux) || os(Windows)
 let platformExcludes = ["Apple", "Mac", "iOS"]
 #else
-let platformExcludes: [String] = []
+let platformExcludes: [String] = excludeAppleSources ? ["Apple", "Mac", "iOS"] : []
 #endif
 
 let isGitHubActions = ProcessInfo.processInfo.environment["GITHUB_ACTIONS"] == "true"
--- a/Sources/SwiftTerm/KittyGraphics.swift
+++ b/Sources/SwiftTerm/KittyGraphics.swift
@@ -5,7 +5,11 @@
 //
 
 import Foundation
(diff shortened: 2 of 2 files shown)


migueldeicaza:
Whoa, this is great!   Happy to merge this change, just a quick follow up question, would love to know how you are using SwiftTerm on Linux.

ASNNetworks:
Thanks for merging!

I'm using SwiftTerm headlessly — a background process owns the PTYs via LocalProcess and keeps a view-less Terminal per session purely as the emulator, as the source of truth for screen state that UI clients attach to and render. So on Linux there's no terminal UI involved at all, just the engine. It has to be a static binary cross-built from a Mac (no Swift toolchain on the target hosts), which is where both of these bit.

It's for a macOS app I'm building, in short: fleet command for terminal AI agents: run a bunch of coding-agent sessions at once, see live status across all of them, and drive any one of them. Sessions live in the daemon rather than the app, which is why the emulator has to run headless. And now I'm working on running it on Linux, so the work doesn't stop when the laptop does and I can connect to it from client devices. Still in closed testing, but hopefully something I can show to users soon when it's in a more mature state :)

migueldeicaza:
Lovely, thank you for letting me know!

One day, I want to have a Gtk+ front-end for Linux machines.

PR #616, opened 2026-08-01
Translate selections when rows move in place

eyelock:
## The problem

A selection is anchored to absolute buffer rows. When a scroll pushes lines into the scrollback that is safe — the rows stay valid and `yDisp` advances, so the selection keeps pointing at its own text.

But several paths move rows *in place*, leaving `yDisp` untouched. Nothing translates the selection anchors, so they end up pointing at whatever text moved into those rows: the highlight sits over the wrong text, and copying returns that wrong text.

| path | effect |
|---|---|
| `scroll()` with `scrollTop != 0` | rows shift up inside a `DECSTBM` region |
| `reverseIndex()` (`ESC M`) | rows shift down inside the region |
| `cmdInsertLines()` (`CSI L`) | rows below the cursor move down |
| `cmdDeleteLines()` (`CSI M`) | rows below the cursor move up |

Full-screen TUIs hit the first one constantly. GitHub Copilot CLI, for example, reserves its tab bar and prompt rows and scrolls the transcript between them:

```
ESC[4;34r      set scroll region to rows 4-34
ESC[34;1H      move to the last row of the region
\r\n           linefeed -> region scrolls up one line
ESC[1;40r      restore the full region
```

Captured from a real session: 254 scroll-region sets and 373 linefeeds, no `CSI S`, no full-screen repaints. Any selection made while output streams silently drifts onto other lines — the highlight stays pinned to the screen while its text scrolls away underneath.

## Reproduction

```swift
let terminal = /* 10 rows */
terminal.feed(text: "\u{1b}[?1049h")
// paint LINE_1 ... LINE_10, one per row

let selection = SelectionService(terminal: terminal)
selection.setSelection(start: Position(col: 0, row: 4), end: Position(col: 10, row: 4))
selection.getSelectedText()   // "LINE_5"

// scroll rows 2...9 up by one, leaving row 1 and row 10 fixed
terminal.feed(text: "\u{1b}[2;9r\u{1b}[9;1H\r\n\u{1b}[1;10r")

selection.getSelectedText()   // "LINE_6"  <- drifted; yDisp is still 0
```

## The change

Selections register with their terminal (held weakly — the views own them) and are translated when rows move in place. Rows outside the affected range keep their position, and a selection that moves out of the range is cleared, since the text it referred to is gone.

Margin mode (`DECSLRM`) shifts only the columns between the margins. A selection cannot be represented as partially shifted, so on those paths an overlapping selection is cleared rather than translated.

Scrolls that go through the scrollback are untouched — the existing behavior there is already correct.

## Tests

`Tests/SwiftTermTests/SelectionScrollTests.swift` adds seven cases: the selection follows its text through a partial-region scroll, a reverse index, an insert-lines and a delete-lines; it is cleared when it moves out of the region; and it is left alone both when anchored outside the region and on a full-screen scroll.

Against `main`, five of the seven fail (for example `"LINE_6" is not equal to "LINE_5"`) and the two controls pass. With this change all seven pass.

The existing suite is unchanged: 66 XCTest and 455 swift-testing cases pass identically before and after.

## Real-world verification

This shipped in a beta of a terminal app built on SwiftTerm and was confirmed by hand: a selection made during streaming GitHub Copilot CLI output now rides up the screen with its text instead of staying pinned, and Cmd+C returns what is highlighted. Claude Code, Codex CLI and Cursor CLI panes are unaffected either way — they render in the normal buffer, where the scrollback path already behaved correctly.

### DIFF
--- a/Sources/SwiftTerm/SelectionService.swift
+++ b/Sources/SwiftTerm/SelectionService.swift
@@ -34,13 +34,23 @@ class SelectionService: CustomDebugStringConvertible {
      * the scrollback, so `yDisp` does not move and the absolute rows the
      * selection is anchored to end up holding different text.
      *
-     * Rows outside the scrolled region keep their position.  A selection that
-     * scrolls out of the region is dropped, since the text it referred to is
-     * gone.
+     * Rows outside the scrolled region keep their position.  A selection is
+     * dropped if it scrolls out of the region or crosses a region boundary.
+     * In those cases, the original text is gone or is no longer contiguous.
      */
     func adjustForInPlaceScroll (top: Int, bottom: Int, lines: Int)
     {
         guard active, lines != 0 else {
+            return
+        }
+
+        let (first, last) = Position.compare (start, end) == .before ? (start, end) : (end, start)
+        let intersectsRegion = first.row <= bottom && last.row >= top
+        guard intersectsRegion else {
+            return
+        }
+        guard first.row >= top && last.row <= bottom else {
+            selectNone ()
             return
         }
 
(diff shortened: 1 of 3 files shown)


migueldeicaza:
This is a great find.   It makes sense. Let me review this

migueldeicaza:
Please test my changes and let me know if this works for your scenarios.

eyelock:
@migueldeicaza thanks for the extra impact fixes I missed, tested and covers my use case.  appreciate the quick response.

PR #617, opened 2026-08-03
Fix dropped child-exit event for fast-exiting processes on macOS

garrettsammons:
## Problem

On macOS, `LocalProcess.startProcessWithForkpty()` creates the child-exit `DispatchSource` and **activates it before installing its event handler**:

```swift
if let cm = childMonitor {
    if #available(macOS 10.12, *) {
        cm.activate()          // armed here...
    } else {
        // Fallback on earlier versions   // ...and never resumed on this path
    }
    cm.setEventHandler(...)     // handler installed only after
}
```

`NOTE_EXIT` is delivered at most once. If the child exits in the window between `activate()` and `setEventHandler()` — routine for a fast-exiting child, especially under CPU load — libdispatch latches the exit event with no handler installed and never redelivers it. `processTerminated()` then never runs, so:

- the child is never `waitpid`-reaped (it lingers as a zombie), and
- `LocalProcessDelegate.processTerminated(_:exitCode:)` is never called, so any code awaiting process exit hangs indefinitely.

The pre-10.12 branch is additionally a no-op: the source is created suspended and that path neither `activate()`s nor `resume()`s it, so the monitor never starts.

## Fix

- Install the event handler **before** activating the source (the ordering libdispatch documents).
- Call `resume()` on the pre-10.12 path.
- Publish `running` / `childfd` / `shellPid` before arming the source, because with the handler installed first, `activate()` can invoke `processTerminated()` synchronously for an already-exited child, and it reads `self.shellPid` (a `0` makes `waitpid(0, …)` target the caller's process group, which never matches the `setsid` child) and `self.childfd`.

No API change and no behavior change for the normal case — this only closes the lost-event race.

## Reproduction / evidence

A standalone reproducer — `forkpty` a child that `_exit()`s immediately, register a `.exit` `DispatchSourceProcess` both ways, and count exit events that never arrive within a timeout — showed **6–60% of exit events lost** with the original ordering under scheduler pressure (a `sched_yield()` between `activate()` and `setEventHandler()` pushed it toward 60%), and **0% lost** after the fix. In an application spawning many short-lived PTY children under load this surfaced as exit-await calls hanging and zombie children accumulating.

Branch is based on current `main`; `swift build` is clean.

### DIFF
--- a/Sources/SwiftTerm/LocalProcess.swift
+++ b/Sources/SwiftTerm/LocalProcess.swift
@@ -511,20 +511,35 @@ public class LocalProcess {
         }
 
         if let (shellPid, childfd) = PseudoTerminalHelpers.fork(andExec: executable, args: shellArgs, env: env, currentDirectory: currentDirectory, desiredWindowSize: &size) {
-#if os(macOS)
-            childMonitor = DispatchSource.makeProcessSource(identifier: shellPid, eventMask: .exit, queue: dispatchQueue)
-            if let cm = childMonitor {
-                if #available(macOS 10.12, *) {
-                    cm.activate()
-                } else {
-                    // Fallback on earlier versions
-                }
-                cm.setEventHandler(handler: { [weak self] in self?.processTerminated () })
-            }
-#endif
+            // Publish process state before arming the exit source below. The
+            // source's event handler (installed just below) can be invoked
+            // synchronously by activate() when the child has already exited,
+            // and processTerminated() reads self.shellPid (a 0 here makes
+            // waitpid(0, ...) target the caller's process group, which never
+            // matches the setsid child) and self.childfd. Setting these first
+            // keeps that early callback correct.
             running = true
             self.childfd = childfd
             self.shellPid = shellPid
+#if os(macOS)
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Thank you for your contribution, apologies it took me this long to review.

Sources/SwiftTerm/LocalProcess.swift
```diff
@@ -511,20 +511,35 @@ public class LocalProcess {
         }
 
         if let (shellPid, childfd) = PseudoTerminalHelpers.fork(andExec: executable, args: shellArgs, env: env, currentDirectory: currentDirectory, desiredWindowSize: &size) {
+            // Publish process state before arming the exit source below. The
+            // source's event handler (installed just below) can be invoked
+            // synchronously by activate() when the child has already exited,
+            // and processTerminated() reads self.shellPid (a 0 here makes
+            // waitpid(0, ...) target the caller's process group, which never
+            // matches the setsid child) and self.childfd. Setting these first
```
migueldeicaza: `processTerminated()` currently reads `self.shellPid`, but it does not read `self.childfd`. Please remove the `childfd` claim, or explain which path uses it, so the rationale stays accurate.

migueldeicaza (commented):
One non-blocking documentation note.

PR #620, opened 2026-08-05
Fix stale rows when repainting in place while scrolled back

evertjr:
## Problem

The terminal's update range is recorded in `buffer.y` space, relative to `yBase`. `updateDisplay` turns it into an invalidation rect as if buffer row `y` were screen row `y`, but `drawTerminalContents` maps screen rects back to buffer rows through `yDisp`:

```swift
let firstRow = displayBuffer.yDisp+Int ((boundsMaxY-dirtyRect.maxY)/cellHeight)
```

Those agree only while the viewport is pinned to the bottom. Scrolled back by `k` rows, a change at buffer row `y` renders at screen row `y + k` while the invalidation still covers screen row `y`, so the rows that actually changed are never repainted. They stay stale until a resize or another scroll forces a full redraw.

Scrolling output masks it: `scroll()` ends with `updateRange(scrollTop)` and `updateRange(scrollBottom)`, so on a full-height screen rows 0 and `rows-1` are both dirty and the `rowEnd == rows - 1` branch invalidates everything anyway. What stays exposed is in-place repainting, where the update range remains a strict subset. Agent TUIs that redraw a status block several times a second (Claude Code, Codex) hit this constantly: scroll up a few lines while one is working and the visible rows stop tracking the buffer.

The alternate screen is unaffected, it has no scrollback so `yDisp == yBase` always.

## Fix

Invalidate the whole view when `yDisp != yBase`.

The pinned-to-bottom case keeps the existing partial rect untouched, including the restricted-region extension from #582. It is less blunt than it reads: `drawTerminalContents` still repaints only the rows intersecting the dirty rect, each read from its correct `yDisp`-relative line.

CoreGraphics path only. Metal builds its own `metalDirtyRange` a few lines below and is not affected.

## Tests

Added a regression test: 20-row view, scrolled back 3 rows, cursor parked on row 2, then a write that neither scrolls nor moves the cursor. Update range is `(2,2)`, and before the fix the invalidated rect is `[256, 288]` while the change renders at `[224, 240]`. Fails on main, passes with the fix. Rest of the suite is green.

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -1851,7 +1851,7 @@ extension TerminalView {
         // drawTerminalContents maps screen rects back to buffer rows via `yDisp`.
         // Those two agree only while the viewport is pinned to the bottom. Once the
         // user scrolls back by `k` rows, the cells that changed are drawn at screen
-        // row `y - k` while the invalidation still covers screen row `y`, so the rows
+        // row `y + k` while the invalidation still covers screen row `y`, so the rows
         // that actually changed are never repainted and keep stale pixels until
         // something forces a full redraw. Invalidate everything in that case; the
         // draw still only repaints rows intersecting the dirty rect and reads each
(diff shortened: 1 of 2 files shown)


migueldeicaza:
This looks good, just a couple of small edits coming in in a second

migueldeicaza:

PR #629, opened 2026-08-11
Fix iOS accessibility reading bounds and coordinates

at-wr:
## Summary
- return NSNotFound, nil, or .zero for invalid reading-content lines instead of indexing outside the display buffer
- clamp current-page ranges to valid visible rows and use the actual final row width
- interpret accessibilityLineNumber(for:) in the view coordinate space UIKit documents
- convert accessibilityFrame(forLineNumber:) results to screen coordinates
- keep reads consistent with the current display buffer and terminal lock
- add platform-neutral boundary tests for line and visible-page calculations

## Rationale
UIKit documents that accessibilityLineNumber(for:) receives a point in the view coordinate space, while accessibilityFrame(forLineNumber:) returns a screen-coordinate frame. TerminalView is a UIScrollView, so converting the input point from the window double-applies its origin and scroll offset. The previous implementation also allowed invalid line indices and constructed page endpoints beyond the final buffer row.

This rebased version targets current main and preserves its render mailbox and accessibility notification behavior.

## Testing
- DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --filter AccessibilityReadingPolicy (5 tests pass)
- DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test (995 tests in 94 suites pass)
- downstream UIKit regression verifies line lookup after a nonzero host-view origin and scroll offset
- strict unsigned iOS build succeeds in the downstream app

### DIFF
--- a/Sources/SwiftTerm/iOS/iOSTerminalView.swift
+++ b/Sources/SwiftTerm/iOS/iOSTerminalView.swift
@@ -3616,121 +3616,155 @@ extension TerminalView: UIAccessibilityReadingContent {
     }
 
     public func accessibilityLineNumber(for point: CGPoint) -> Int {
-        return Int(floor(max(point.y,0) / cellDimension.height))
-    }
-    
-    func startingLine(forLineNumber lineNumber: Int) -> Int {
-        withTerminal { _ in
-            startingLineLocked(forLineNumber: lineNumber)
-        }
+        let contentY = Double(point.y)
+        let lineHeight = Double(cellDimension.height)
+        return withTerminal { terminal in
+            AccessibilityReadingPolicy.lineNumber(
+                atContentY: contentY,
+                lineHeight: lineHeight,
+                lineCount: terminal.displayBuffer.lines.count
+            )
+        } ?? NSNotFound
     }
 
-    func startingLineLocked(forLineNumber lineNumber: Int) -> Int {
+    private func accessibilityLogicalLineRangeLocked(
+        containing lineNumber: Int
+    ) -> ClosedRange<Int>? {
         terminal.terminalLock.preconditionLocked()
+        let buffer = terminal.displayBuffer
+        guard buffer.lines.indices.contains(lineNumber) else {
+            return nil
+        }
+
         var startingLine = lineNumber
         while startingLine >= 1 {
             startingLine -= 1
-            if terminal.buffer.lines[startingLine + 1].isWrapped {
+            if buffer.lines[startingLine + 1].isWrapped {
                 continue
             }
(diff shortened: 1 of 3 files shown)


Sources/SwiftTerm/iOS/iOSTerminalView.swift
```diff
@@ -3367,99 +3367,115 @@ extension TerminalView: UIAccessibilityReadingContent {
     }
 
     public func accessibilityLineNumber(for point: CGPoint) -> Int {
-        return Int(floor(max(point.y,0) / cellDimension.height))
+        // UIKit supplies this point in screen coordinates. Converting through
+        // the scroll view also accounts for its bounds origin/content offset.
+        let localPoint = convert(point, from: nil)
+        return AccessibilityReadingPolicy.lineNumber(
```
migueldeicaza: I think that accessibilityLineNumber receives screen coordinates, and the call to convert does it from window coordinates, I think this needs reviewing.

From their docs:

https://developer.apple.com/documentation/uikit/uiaccessibilityreadingcontent/accessibilitylinenumber%28for%3A%29

> A point within the bounds of the receiver’s view space, in screen coordinates. That is, a point for which [self pointInside:point withEvent:event] == YES.

The "nil" to convert indicates window coordinates.

migueldeicaza (changes requested):
Thank you for sending this along!

Sources/SwiftTerm/iOS/iOSTerminalView.swift
at-wr: Confirmed and fixed in f66ad7a. The iOS 26.5 SDK UIAccessibility.h contract says the point is in the view coordinate space. A Reach UIKit integration regression positions the terminal away from the window origin, scrolls retained history, derives the local point from accessibilityFrameForLineNumber, and proves the old conversion maps requested line 3 to line 54. The implementation now consumes point.y directly; the same regression passes with the fix. The v1.18 fork suite passes 665 tests and current main passes 701 tests on rerun.

at-wr:
Rebased this onto current main as a single commit (4b918d7) and incorporated the coordinate-space review. UIKit documents the input point in the view coordinate space, so the new version uses point.y directly; only the returned line frame is converted to screen coordinates. I also left the newer render mailbox and accessibility notification flow unchanged. The full current suite passes: 995 tests in 94 suites.

migueldeicaza:
Oops I just saw the close, and missed your earlier message where you fixed the issue.  Was the bug fixed?

PR #632, opened 2026-08-13
Guard the unknown-DECRQSS log against a non-ASCII payload

faisalmumtaz89:
The DCS `$q` (DECRQSS) handler decodes the request payload as ASCII, which is `nil` for any byte ≥ 0x80. The unknown-request branch then logs it as `\(newData!)`, so a DECRQSS whose payload carries a non-ASCII byte traps on the force-unwrap:

```
ESC P $ q <0xA0> ESC \   →   Fatal error: Unexpectedly found nil while unwrapping an Optional value
```

This is reachable from ordinary terminal output. The fix logs `newData ?? ""` instead — the nil-safe form already used for the response in this same handler, and everywhere else this file decodes ASCII (`if let` / `guard let` / `?? default`).

Adds a `DcsTests` regression that feeds a non-ASCII DECRQSS payload and confirms it no longer crashes.

### DIFF
--- a/Sources/SwiftTerm/Terminal.swift
+++ b/Sources/SwiftTerm/Terminal.swift
@@ -1119,7 +1119,7 @@ open class Terminal {
             default:
                 ok = 0 // this means the request is not valid, report that to the host.
                 // invalid: DCS 0 $ r Pt ST (xterm)
-                terminal.log ("Unknown DCS + \(newData!)")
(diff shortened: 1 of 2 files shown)


migueldeicaza:
Oops, thank you!   I really need to kill that bad habit that I had early on using "foo!"

PR #641, opened 2026-08-18
perf: bound terminal output scheduling

dliu120:
## Why

Large PTY reads can monopolize the main delivery queue, while interactive output and hidden terminal views can continue presentation work in the same main-loop turn. Both reduce the time available for input handling during heavy terminal output.

## What changed

- Deliver queued PTY output in ordered slices of at most 16 KiB while preserving the existing read size and backpressure thresholds.
- Coalesce interactive display requests through the existing pending-display gate, including feeds already running on the main thread.
- Add an explicit presentation-active state: parsing continues while hidden, but display, accessibility, Metal, cursor/text blink, and related callbacks pause.
- Generation-gate queued presentation callbacks so stale work cannot resume after visibility transitions.
- Mark the full screen dirty and schedule one refresh when presentation resumes.
- Preserve synchronized-output behavior.
- Remove a duplicate debug-display update from the display wrapper.
- Add focused coverage for byte ordering, delivery bounds, queue compaction, interactive display coalescing, and presentation suspension/resume.

## Testing

- `swift test --filter LocalProcessSchedulingTests`
- focused presentation/scheduling coverage: 7/7
- `swift test` (701 tests across 60 suites)
- `git diff --check`
- [ ] CI green

## Notes

This reduces and suspends bounded presentation work but does not claim a measured end-to-end latency improvement without runtime traces.

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -2498,7 +2498,6 @@ extension TerminalView {
     {
         updateTextBlinkLifecycle()
         updateDisplay (notifyAccessibility: true)
-        updateDebugDisplay()
         pendingDisplay = false
     }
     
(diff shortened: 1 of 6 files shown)


migueldeicaza:
I redid the IO system in SwiftTerm in this branch, which will soon become the main branch (v2), please take it for a spin as it touches the same issues:

new-io-perf-glyph-cache

migueldeicaza:
Please reopen this PR if the issue was not fixed in main, but I changed the entire IO pipeline to address this and other issues on main.

PR #646, opened 2026-08-20
Make the OSC 9;4 progress bar visible and smooth

stuartcarnie:
The progress bar added in `13aff25` never appeared when the Metal renderer was on, and stuttered when it did appear. Running a script that emits OSC 9;4 showed nothing at all in an app that defaults to Metal.

There were three separate causes.

**The bar was hidden behind the Metal surface.** `insertMetalView` inserts the Metal view below the caret. The caret sits above the progress bar, so the opaque, full-bounds Metal surface covered the bar completely, and only the scroller was raised back afterwards. `setupProgressBar` also added the bar below the caret that `setupOptions` had already installed. The bar is now added topmost and raised again after any renderer swap.

**The pulse animation restarted on every report.** Applications repeat `9;4;3` continuously, often ten times a second. Each report tore down the slide animation and started a new one, so the bar juddered in place instead of sweeping. It now keeps running until the bounds or the state change.

**The percentage bar jerked forward and back.** Setting `barLayer.frame` inside a `CATransaction` creates an implicit animation that starts from the layer's previous model value, not from where the bar currently is. With reports arriving faster than the animation lasts, each one snapped the bar forward and eased again. Restarting an ease-in-out at that rate also drops the bar to zero velocity many times a second. The bar layer is now anchored at its leading edge, and the width is carried by an additive animation, so animations still in flight sum with each new one and the motion stays continuous. This is what SwiftUI does when it retargets, which is why Ghostty's bar looks smooth.

Two smaller changes came with it. The per-view progress retention is gone, replaced by Ghostty's rule applied in the bar view: an explicit percentage wins, a paused report with no percentage reads as complete, and everything else is indeterminate. Bar thickness is now one shared constant used by both hosts instead of a literal in each.

**Testing.** The full suite passes. A new test in `MetalToggleTests` pins the bar as the topmost subview, since that is the failure that made the feature invisible. I also checked the parser against the exact bytes cargo emits, including the `ESC \` terminator and the `9;4;0;` clear form, and measured the bar's pixel width over time through a real PTY with Metal on: the width increases monotonically and the speed holds steady within sampling noise.

### DIFF
--- a/Sources/SwiftTerm/Apple/TerminalProgressBarView.swift
+++ b/Sources/SwiftTerm/Apple/TerminalProgressBarView.swift
@@ -21,12 +21,18 @@ final class TerminalProgressBarView: ProgressBarBaseView {
     private let barLayer = CALayer()
     private let indeterminateAnimationKey = "terminalProgressIndeterminate"
 
+    /// Thickness of the bar, in points. Ghostty uses 2; one more point reads
+    /// better against terminal content without stealing a row.
+    static let preferredHeight: CGFloat = 3
+
     private let barWidthRatio: CGFloat = 0.25
     private let indeterminateDuration: CFTimeInterval = 1.2
     private let determinateDuration: CFTimeInterval = 0.2
 
     private var state: Terminal.ProgressReportState = .remove
     private var progress: UInt8?
+    /// Bounds the running indeterminate animation was built for.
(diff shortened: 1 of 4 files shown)


migueldeicaza:
Thank you so much for digging into all these issues.  Will merge and review asynchronously

PR #650, opened 2026-08-21
Fix implicit-link boundaries and link cursor flicker

furgalep:
## Summary
- stop including end-of-line padding spaces in implicit filesystem-path matches
- make AppKit cursor updates preserve the pointing hand over visible Command-links instead of resetting to the I-beam
- perform one lock-protected link lookup per cursor update and skip implicit matching when it cannot be visible
- add focused regressions

Terminal emulators pad every row with blank cells. Treating those cells as part of a link makes the hover underline and click target extend to the right edge. Separately, SwiftTerm's unconditional I-beam in `cursorUpdate(with:)` competed with host link-hover handling and caused rapid cursor flicker.

The application-specific ` at ` path boundary has been removed. A host-provided link resolver/policy would be a better follow-up for cases that need filesystem- or session-specific validation, since SwiftTerm may display remote filesystems where local probing is not authoritative.

## Tests
- `swift test --filter GhosttyImplicitLinkDetectionTests`
- `swift test --filter MacDefaultLinkTests`
- `swift test` (986 tests passed)

### DIFF
--- a/Sources/SwiftTerm/Mac/MacTerminalView.swift
+++ b/Sources/SwiftTerm/Mac/MacTerminalView.swift
@@ -1660,7 +1660,30 @@ open class TerminalView: NSView, NSUserInterfaceValidations, TerminalDelegate {
     
     public override func cursorUpdate(with event: NSEvent)
     {
-        NSCursor.iBeam.set ()
+        let hit = calculateMouseHit(with: event).grid
+        let hasCommandModifier = commandActive || event.modifierFlags.contains(.command)
+        linkCursor(at: hit, hasCommandModifier: hasCommandModifier).set()
+    }
+
+    func linkCursor(at position: Position, hasCommandModifier: Bool) -> NSCursor
+    {
+        let match: Terminal.LinkMatch?
+        switch linkHighlightMode {
+        case .hover, .hoverWithModifier:
+            // Reuse the hover lookup so cursor updates do not run the implicit-link
+            // regular expression twice.
+            match = updateHoverLink(at: position, commandOverride: hasCommandModifier)
+        case .always, .alwaysWithModifier:
+            // Implicit links are never visible in these modes.
+            match = withTerminal { terminal in
+                terminal.linkMatch(at: .buffer(position), mode: .explicitOnly)
+            }
+        }
+        guard let match,
+              linkVisibleForClick(match: match, hasCommandModifier: hasCommandModifier) else {
+            return .iBeam
+        }
+        return .pointingHand
     }
     
     func makeFirstResponder ()
@@ -3637,7 +3660,11 @@ open class TerminalView: NSView, NSUserInterfaceValidations, TerminalDelegate {
         }
     }
 
-    func updateHoverLink(at position: Position, commandOverride: Bool? = nil)
+    @discardableResult
+    func updateHoverLink(
+        at position: Position,
+        commandOverride: Bool? = nil
+    ) -> Terminal.LinkMatch?
     {
         let hoverModes: [LinkHighlightMode] = [.hover, .hoverWithModifier]
         guard hoverModes.contains(linkHighlightMode) else {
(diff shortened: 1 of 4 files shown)


migueldeicaza:
The issue with the "at" is that it sounds like it might be a valid file reference, and I can imagine other strings that we could argue whether they should be part of it or not.   I wonder if we do need a different way of detecting these links.

Sources/SwiftTerm/Mac/MacTerminalView.swift
```diff
@@ -1654,13 +1043,24 @@ open class TerminalView: NSView, NSUserInterfaceValidations, TerminalDelegate {
...
-        NSCursor.iBeam.set ()
+        let hit = calculateMouseHit(with: event).grid
+        let hasCommandModifier = commandActive || event.modifierFlags.contains(.command)
+        updateHoverLink(at: hit, commandOverride: hasCommandModifier)
+        linkCursor(at: hit, hasCommandModifier: hasCommandModifier).set()
+    }
+
+    func linkCursor(at position: Position, hasCommandModifier: Bool) -> NSCursor
+    {
+        guard let match = terminal.linkMatch(at: .buffer(position), mode: .explicitAndImplicit),
```
migueldeicaza: This needs to use withTerminal, as we are now multi-threaded.   I am

migueldeicaza: Additionally, the same call always uses .explicitAndImplicit. With the default .hoverWithModifier mode, it runs the full implicit-link regex even when Command is not pressed. When Command is pressed, updateHoverLink performs the lookup first, and linkCursor immediately performs it again. Use implicitLinkCouldBeVisible to select the mode, or reuse the result from updateHoverLink.

migueldeicaza (changes requested):
Thank you for taking the time to send these along - they look wonderful.

I am still mulling over the regex change, and whether we should do something to make this configurable instead or extend the probing infrastructure - perhaps we should validate the existence (it would only work for local mode)

furgalep:
Addressed the review feedback and force-pushed the rebased branch:

- rebased onto current `upstream/main`
- wrapped link matching in `withTerminal`
- cursor updates now reuse the hover lookup, so the implicit regex runs at most once
- `.hoverWithModifier` without Command exits before any implicit lookup; always modes use `.explicitOnly`
- removed the application-specific ` at ` boundary heuristic

I agree that filesystem/session-specific validation belongs behind a configurable host resolver or policy rather than in the generic regex. Local probing would not be authoritative for remote/container/tmux sessions, so I left that for a separate API discussion.

Validation: focused link suites and the full `swift test` run pass (986 tests).

Sources/SwiftTerm/Mac/MacTerminalView.swift
```diff
@@ -1660,7 +1660,30 @@ open class TerminalView: NSView, NSUserInterfaceValidations, TerminalDelegate {
     
     public override func cursorUpdate(with event: NSEvent)
     {
-        NSCursor.iBeam.set ()
+        let hit = calculateMouseHit(with: event).grid
+        let hasCommandModifier = commandActive || event.modifierFlags.contains(.command)
+        linkCursor(at: hit, hasCommandModifier: hasCommandModifier).set()
+    }
+
```
migueldeicaza, line 1667: AppKit calls cursorUpdate(with:) when the pointer enters a cursor rectangle. 

resetCursorRects() defines one rectangle for the entire view. Moving from normal text to a link therefore updates the underline but not the cursor. Pressing or releasing Command while stationary also leaves a stale cursor. Update the cursor from mouseMoved and flagsChanged, or invalidate the cursor rectangles when link state changes. The new test calls linkCursor directly, so it does not test this event-routing behavior.

migueldeicaza (commented):
Thank you for the follow up!   One small little detail.

Issue #647, opened 2026-08-20
Downstream fixes we can prepare as upstream PRs


everlof:
Hi Miguel,

We maintain a SwiftTerm fork for Threading, a terminal host for macOS and iOS. We recently reconciled the fork with v1.20.0 and sorted out which downstream changes look reusable.

Before we spend time rebasing them onto current `main`, would you be interested in any of the areas below? We would send each accepted area as a small PR with its tests. The [full v1.20.0 comparison](https://github.com/migueldeicaza/SwiftTerm/compare/v1.20.0...everlof:main) is available, but it is an inventory rather than a ready-to-merge PR. Current `main` has changed substantially since v1.20.0, especially around I/O, so we would adapt each patch to the current design.

### 1. `LocalProcess` lifecycle and final-output ordering

Our fork:

- reaps the exact child PID and keeps the instance unavailable for relaunch until reaping finishes;
- waits for both child exit and PTY EOF before sending the termination callback after a natural exit;
- prevents queued output from an old process generation from reaching a newly launched process;
- reaps children after `terminate()` and `deinit`, rather than cancelling the exit monitor and leaving a zombie; and
- bounds queued main-thread output with high and low water marks so a fast producer eventually blocks on the PTY instead of growing memory without limit.

There are nine tests covering instant exits, rapid relaunch, concurrent processes, termination, deallocation, stale output, and sustained backpressure. The final-output part appears to address #370.

The new `TerminalIOPipeline` on `main` already handles backpressure differently. We would port the lifecycle and callback-ordering parts that still apply, not restore the old reader.

### 2. Incremental recent-buffer text

We added a bounded logical-text read for hosts that scan terminal output for paths, links, or transcript updates. It joins soft-wrapped rows, preserves hard line breaks, omits an incomplete oldest logical line, and never returns more than the requested UTF-8 budget.

The incremental form returns an absolute row cursor. A later read skips stable scrollback but always reads the visible screen again, since a full-screen application can repaint those rows in place. A stale cursor after a reset falls back to a complete bounded read.

The implementation and focused tests are in [fdebb7b](https://github.com/everlof/SwiftTerm/commit/fdebb7b44d71fa4cd9f13942dd7c968273dd5674). For current `main`, this may fit better as a concurrency-safe `TerminalView` snapshot API than as the exact v1.20 `Terminal` method.

### 3. Avoid reflowing hidden scrollback during alternate-screen resize

[...]

PR #651, opened 2026-08-24
Add color-scheme update notifications (DEC mode 2031)

evertjr:
Implements the color-palette-update-notifications extension ([Contour spec](https://github.com/contour-terminal/contour/blob/master/docs/vt-extensions/color-palette-update-notifications.md)), which Ghostty, kitty, WezTerm and Contour already ship. Theme-aware TUIs use it to repaint when the terminal switches between light and dark at runtime; without it, apps like OpenCode and cursor-agent keep painting with the palette they sampled at boot, which can leave them unreadable after a system appearance change.

SCR-20260824-jwde

SCR-20260824-jwje

**Emulator side (`Terminal.swift`)**
- `DECSET`/`DECRST` `?2031` toggles the subscription (`colorSchemeUpdatesEnabled`), cleared by a full reset.
- `DECRQM ?2031` reports the subscription state.
- `CSI ?996 n` reports the current preference as `CSI ?997;1n` (dark) or `CSI ?997;2n` (light).
- New `TerminalColorScheme` enum and a read-only `colorScheme` property.

**Host API**
- `updateColorScheme(_:notify:)` records the preference and, by default, sends `CSI ?997 n` to a subscribed application. Call it after installing a new palette so the application's follow-up OSC 10/11 query already sees the new colors.
- `notifyColorScheme()` re-sends the current preference. `notify: false` plus this call lets a host record the scheme immediately (so `?996 n` and `DECRQM` answer correctly) while deferring the announcement.

**Why the split matters for hosts.** Applications answer the notification by re-querying OSC 10/11 with a short timeout (cursor-agent 100ms, OpenTUI 250ms) and ignore further notifications while that query is pending. Since the reply is produced on the host's dispatch queue, a host that sends the notification while it is busy re-rendering after an appearance flip can strand the app on the old theme. Deferring the notification until the queue is responsive, and optionally re-sending once after those timeouts, avoids it. That timing policy is deliberately left to hosts; the library only exposes the two calls.

No view-layer changes. The emulation is platform-independent and available to the macOS, iOS and headless terminals alike; the `colorScheme` default is documented so hosts seed it once the initial palette is installed.

Tests: `ColorSchemeNotificationTests` covers the query, subscribe/unsubscribe, DECRQM, full reset, and the record-without-notify path (7 tests).

Related: #647 lists this extension as something another downstream implemented too; happy to reconcile if their shape differs.

### DIFF
--- a/Sources/SwiftTerm/Terminal.swift
+++ b/Sources/SwiftTerm/Terminal.swift
@@ -10,6 +10,15 @@
 
 import Foundation
 
+/// The light/dark preference represented by the terminal's current color palette.
+///
+/// This is reported to applications through the color-scheme notification protocol:
+/// https://github.com/contour-terminal/contour/blob/master/docs/vt-extensions/color-palette-update-notifications.md
+public enum TerminalColorScheme: Equatable {
+    case dark
+    case light
+}
+
 /**
  * The terminal delegate is a protocol that must be implemented by a class
  * that would provide a user interface for the terminal, and it is used by the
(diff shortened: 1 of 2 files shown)


migueldeicaza:

PR #652, opened 2026-08-25
Expose color-scheme updates through TerminalView

everlof:
#651 added color-scheme state and notification methods to `Terminal`. Apps embedding SwiftTerm's AppKit or UIKit views cannot safely call those methods directly: the views do not expose the mutable terminal, and their `withTerminal` helper is internal.

This adds `updateColorScheme(_:notify:)` and `notifyColorScheme()` to the shared `TerminalView` extension. Both calls go through `TerminalRenderOwner` and the terminal lock, following the same boundary used by the other public view commands. Palette installation and notification timing remain host decisions.

Verification:

- `swift test` (998 tests)
- `swift build --target SwiftTerm -Xswiftc -strict-concurrency=complete`
- generic iOS sample build


migueldeicaza:

PR #653, opened 2026-08-25
Keep LocalProcess alive until its exact child is reaped

everlof:
Follow-up to #647.

`LocalProcess.terminate()` currently clears process state and cancels its exit source before the child has been reaped. That allows a relaunch to begin while the previous PID is still live, and the nonblocking `waitpid` in the exit callback can miss the exact child.

This keeps the session active after `SIGTERM` until a blocking `waitpid` has reaped that PID. Sends are rejected once termination starts, relaunch remains blocked until cleanup completes, exit status is normalized before delivery, and deinitialization transfers the PID to a detached reaper so releasing `LocalProcess` cannot leave a zombie. The existing I/O pipeline shutdown behavior is unchanged.

Verification:

- `swift test`

### DIFF
--- a/Sources/SwiftTerm/LocalProcess.swift
+++ b/Sources/SwiftTerm/LocalProcess.swift
@@ -27,6 +27,7 @@ private struct LocalProcessSessionState {
     var childfd: Int32 = -1
     var shellPid: pid_t = 0
     var running = false
+    var terminationRequested = false
     var pipeline: TerminalIOPipeline?
     var writeChannel: DispatchIO?
     var loggingDirectory: String?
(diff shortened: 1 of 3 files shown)


migueldeicaza:
I have been trying to fix this, as each fix uncovers another scenario that we did not take into consideration, I might have to start this from scratch :-(

migueldeicaza:
Ok, after several passes, I ended up with a different approach.  

Fixed on main.

migueldeicaza:
I had to land even more fixes they are all now in.

everlof:
Amazing, thank you very much for your efforts!

PR #655, opened 2026-08-25
Defer hidden normal-buffer reflow while alternate screen is active

everlof:
Follow-up to #647.

Every resize currently reflows both terminal buffers. When a full-screen application owns the alternate screen, that means each live-resize update can reflow thousands of hidden normal-buffer scrollback rows that cannot be displayed.

This keeps the alternate buffer at the live grid size but leaves the hidden normal buffer at its previous dimensions. When the terminal returns to the normal screen, SwiftTerm reflows it once at the final grid size and updates its saved cursor and tab stops through the same resize rules used by the immediate path.

The regression coverage exercises repeated alternate-screen resizes, preservation of scrollback content and tab stops, and parity with a visible normal-buffer resize.

Verification:

- `swift test`

### DIFF
--- a/Sources/SwiftTerm/Terminal.swift
+++ b/Sources/SwiftTerm/Terminal.swift
@@ -1092,6 +1092,12 @@ open class Terminal {
             return
         }
         semanticNoteAlternateScreenSwitch()
+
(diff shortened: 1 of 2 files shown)


migueldeicaza:
Thank you!   Brilliant improvement!

PR #656, opened 2026-08-25
Preserve valid selections while output is streaming

everlof:
Follow-up to #647.

`TerminalView` currently clears an active selection before every feed whenever mouse reporting is allowed. That makes ordinary process output erase a manual selection even when its buffer coordinates are still valid.

The terminal core already registers each `SelectionService`, moves its anchors when rows move, and clears it when a scroll operation can no longer represent the selected text. This change lets that existing buffer mutation logic decide whether the selection survives. Search results are still invalidated before parsing.

The same rule now applies to regular feeds, borrowed local-process feeds, and the deprecated line-feed delegate hooks on both Apple views.

Verification:

- `swift test`
- generic iOS device build with code signing disabled

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -1220,17 +1220,12 @@ extension TerminalView {
     func configureFeedSender() {
         let owner = renderOwner
         let signal = frameSignal
-        let crossThreadState = crossThreadState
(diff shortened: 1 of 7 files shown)


migueldeicaza:
Ok this is a lovely feature, but it is missing a few things, let me rework this.

It does not handle some scenarios like buffer switching.

migueldeicaza:
Landed #661 inspired by this.

PR #657, opened 2026-08-25
Bound Apple wheel reports and improve iOS gesture routing

everlof:
Follow-up to #647, covering the wheel half of item 5 and the iOS routing in item 6.

AppKit can fold acceleration into a large wheel delta. Forwarding every unit as a terminal mouse report can overwhelm a slow full-screen application. Classic wheel events now send one report per event. Precise trackpad and iOS drag input share a 100 reports per second token bucket with an immediate burst of six. Holding Option on macOS bypasses mouse reporting and uses local scrollback; in an alternate buffer, where SwiftTerm has no local scrollback, the gesture does nothing.

On iOS, a mouse-tracking application can receive a tap even when the software keyboard is dismissed. One-finger vertical drags become bounded wheel reports, while two fingers remain available for local scrollback. Alternate Scroll Mode maps the one-finger gesture to cursor keys when mouse tracking is off, and `allowMouseReporting` consistently returns the gesture to UIKit when disabled. Tap forwarding remains available to embedders through `forwardTap(at:)`; wheel-drag accumulation stays owned by SwiftTerm's gesture lifecycle.

Verification:

- focused wheel and alternate-scroll suite: 19 tests
- generic iOS device build with code signing disabled

The complete local suite reaches 1,070 passing tests. `KittyTransmissionTests/testKittySharedMemoryLoad` fails independently with `EINVAL: bad payload`; none of the Kitty implementation or test files differ in this PR.


Sources/SwiftTerm/Mac/MacTerminalView.swift
```diff
@@ -3801,6 +3802,7 @@ open class TerminalView: NSView, NSUserInterfaceValidations, TerminalDelegate {
         let scrollRouting = withTerminal { terminal in
             let reportsMouse = allowMouseReporting &&
                 !shiftBypassesMouseReportingLocked(for: event) &&
+                !event.modifierFlags.contains(.option) &&
```
migueldeicaza: This check has some unintended side effect.

If the app is in alternate mode, with mouse tracking, and it enables alternate scroll mode, if the user scrolls while holding options.

Because this is false, later, the code falls in here:

```
} else if scrollRouting.isAlternate {
    // Sends cursor keys
}
```

So option-wheel gesture intended for local handling can perform an action inside fu;l; screen applications.

Unclear what we should do in this case since we do not have a scroll back on the alternate buffer.   I think the least surprising thing would be to do nothing, rather than sending the keys, and if one day we do alternate buffer scrolling (I think there is a feature request for this), we could wire it up there.

Sources/SwiftTerm/iOS/iOSTerminalView.swift
```diff
@@ -1049,37 +1084,87 @@ open class TerminalView: UIScrollView, UITextInputTraits, UIKeyInput, UIScrollVi
...
+            forwardWheelDrag(distance: travelled, gestureRecognizer: gestureRecognizer)
+        default:
+            break
+        }
+    }
+
+    /// Reports whole-cell drag distance as bounded wheel presses, or as cursor keys when an
+    /// alternate-screen application relies on Alternate Scroll Mode instead of mouse tracking.
+    public func forwardWheelDrag(
+        distance: CGFloat,
```
migueldeicaza, line 1149: I am not sure why this is public, but it carries one problem if it is - since the contract is no longer internal, and it is exposed publicly. 

Imagine two gestures:

1. Host calls forwardWheelDrag(distance: 15)
wheelDragAccumulator becomes 15
15 / 20 = 0 lines
Method returns
Gesture ends

2. Host starts a new gesture
Host calls forwardWheelDrag(distance: 5)
wheelDragAccumulator becomes 20
20 / 20 = 1 line
A wheel report is sent

The second gesture moved only 5 points, but it sends a wheel report because it inherited 15 points from the previous gesture.

Our internal system already resets the value in the pan handler, but not on this one.

This violates the gesture boundary: each drag should start with clean distance state. A suitable fix is to handle .began inside the public forwarding path, and have the internal handler forward all gesture states. Another option is a public beginWheelDrag() operation.   

So if we need this to be public, we need a way of addressing that.

migueldeicaza (changes requested):
Fabulous changes, some comments.

Sources/SwiftTerm/Mac/MacTerminalView.swift
everlof: Good catch. I replaced the boolean fallthrough with an explicit four-way route: mouse, cursor keys, local scrollback, or none. Option and uncaptured Shift now use local scrollback in the normal buffer and do nothing in the alternate buffer, so neither can become cursor-key input. The suppressed route also clears partial trackpad accumulation. I added coverage for both modifiers in an alternate buffer with mouse tracking enabled.

Sources/SwiftTerm/iOS/iOSTerminalView.swift
everlof: You are right. I checked the downstream production call sites and we do not invoke forwardWheelDrag outside SwiftTerm; only tests called it directly. I removed its public exposure instead of publishing an unnecessary lifecycle contract. The built-in recognizer remains the owner, now resets fractional distance at begin and at every terminal gesture state, and has a focused two-gesture regression for the carry-over case.

everlof:
CI note after the rebase: the exact base commit, `9c2518e`, has the same job matrix in [its Swift run](https://github.com/migueldeicaza/SwiftTerm/actions/runs/33117650486): `build`, `swift6-tools`, `linux-headless`, and `thread-sanitizer` fail, while `swiftpm-strict-concurrency` passes.

migueldeicaza:
Thank you - I had Codex/AI review this round, would you mind validating these ones?

[P2] iOS can forward suppressed fractional movement. [forwardWheelDrag](https://github.com/migueldeicaza/SwiftTerm/blob/037f1d0e6a80e958fbd602a1b9fb88dbcc0a631c/Sources/SwiftTerm/iOS/iOSTerminalView.swift#L1122-L1175) updates the accumulator before it determines the route. If a Shift-bypassed drag moves 15/20 of a cell, then Shift is released and the drag moves another 5/20, the application receives a wheel report that includes the suppressed movement. Determine the route first and reset the accumulator for .none, or reset it when the route changes.

[P2] macOS shares fractional movement across different routes. [scrollAccumulator](https://github.com/migueldeicaza/SwiftTerm/blob/037f1d0e6a80e958fbd602a1b9fb88dbcc0a631c/Sources/SwiftTerm/Mac/MacTerminalView.swift#L3808-L3851) is not reset when the route changes. A sub-cell mouse-reporting event can therefore combine with a later Option-local event and unexpectedly scroll locally, or the reverse. Reset the accumulator when WheelRoute changes, or keep separate accumulators.

[P3] allowMouseReporting documents the old behavior. Its [public comment](https://github.com/migueldeicaza/SwiftTerm/blob/037f1d0e6a80e958fbd602a1b9fb88dbcc0a631c/Sources/SwiftTerm/iOS/iOSTerminalView.swift#L137-L148) says disabling reporting converts pans to cursor keys. The new implementation instead restores UIKit scrolling. Update the API documentation.

everlof:
Thanks, all three were right. Fixed in e0575bd.

**iOS forwards suppressed fractional movement.** Confirmed. `forwardWheelDrag` now determines the route first; a `.none` route resets the accumulator and returns, so travel a Shift bypass suppressed is dropped rather than carried into the first report once Shift lifts.

**macOS shares fractional movement across routes.** Confirmed. The accumulator is now `WheelDistanceAccumulator<Route>` (renamed, since macOS uses it too): it remembers the route it banked for and starts over when the route changes. Two new view tests drive a sub-cell mouse-reporting event followed by an Option event, and the reverse; both fail without the reset.

**`allowMouseReporting` documents the old behaviour.** Updated. The comment now says disabling it returns one-finger drags to UIKit's own scrolling and generates no cursor keys in their place.

Focused suite is 25 tests, generic iOS build and `-strict-concurrency=complete` are clean.

migueldeicaza:
Thank you!

PR #659, opened 2026-08-26
Report hasText as true so the delete key keeps its auto-repeat

cunderw:
### The problem

`TerminalView.hasText` reports `false` whenever the input-session shadow is empty:

```swift
public var hasText: Bool {
    return !textInputStorage.isEmpty
}
```

`textInputStorage` is cleared on every Enter and never holds anything the far side echoed, so at an ordinary shell prompt it is empty — even with a full command line on screen.

`hasText` is a `UIKeyInput` requirement, and the system uses it to decide whether the delete key has anything to act on. Reporting `false` is what stops **key repeat: holding backspace deletes one character and stops**, on the software keyboard and on a hardware one.

### Why this looks like an inconsistency rather than a judgement call

`deleteBackward()` already treats the empty-storage case as "send a backspace to the terminal", and says so in its own comment:

```swift
if rangeStartIndex == 0 {
    // This is the case when the user hits backspace, but there is no text in the
    // text input buffer.  This happens for example when text has been pasted.
    // In that scenario, we should just send the backspace character to the terminal
    pendingAutoPeriodDeleteWasSpace = false
    self.sendBackspaceKey()
```

So in exactly the state where `hasText` answers "nothing to delete", `deleteBackward()` deletes anyway. The view tells the system it has no text and then acts as though it does — and the only visible consequence is the lost repeat.

For a terminal the honest answer is that there is always something to delete: what a backspace *means* is the remote's decision, not the view's.

### The change

```diff
     public var hasText: Bool {
-        return !textInputStorage.isEmpty
+        // A terminal always has something a backspace could act on: what it means is the
+        // remote's decision, not this view's. Reporting the input-session shadow here made
+        // `hasText` false at a shell prompt — where `deleteBackward()` sends a backspace
+        // regardless — which cost the delete key its key repeat.
+        return true
     }
```

### Reproducing

Any `TerminalView` attached to a shell. Type a few characters at a prompt, then press and hold backspace: one character is deleted. Tapping repeatedly deletes normally, so it is specifically the repeat that is missing. `hasText` returns `false` throughout.

### If you would rather not change the default

Marking it `open` instead would let a subclass answer for itself and is enough to unblock downstream use. It is the smaller change, though it leaves the inconsistency above in place for every other embedder. Note `deleteBackward()` is already `open` while `hasText` is `public`, so the pair cannot currently be overridden together.

Happy to take this either way, or to add a test if there is a pattern you would like it to follow.

---

Found while building [Drover](https://github.com/cunderw/drover), an iOS SSH client that embeds SwiftTerm. Verified against `main` at b202ea5 and against v1.18.0, which is the version Drover pins.

### DIFF
--- a/Sources/SwiftTerm/iOS/iOSTerminalView.swift
+++ b/Sources/SwiftTerm/iOS/iOSTerminalView.swift
@@ -2040,7 +2040,15 @@ open class TerminalView: UIScrollView, UITextInputTraits, UIKeyInput, UIScrollVi
     }
     
     public var hasText: Bool {
-        return !textInputStorage.isEmpty
+        // A terminal always has something a backspace could act on: what a backspace
+        // *means* is the remote's decision, not this view's. Answering from
+        // `textInputStorage` — which is cleared on every Enter and never holds what the
+        // far side echoed — made this `false` at an ordinary shell prompt, where
+        // `deleteBackward()` sends a backspace regardless (see its own comment there).
+        // The system reads `hasText` to decide whether the delete key has anything to
+        // act on, so reporting `false` cost that key its auto-repeat: holding backspace
+        // deleted one character and stopped.
+        return true
     }
 
     func isAutoPeriodReplacement(_ text: String) -> Bool {


migueldeicaza:
Mhm, I am concerned, because this was changed here: https://github.com/migueldeicaza/SwiftTerm/commit/c1b7251c9a5d7736021a3c1142f33b54b7d6a493

I need to test this, but a quick hack would be to make this open for now, can you submit that change, while I test this other scenario?

migueldeicaza:
One thing that might be worth doing here is this: as you point out hasText does not have much of a say as to how the remote end might interpret the input, but luckily, we now have OSC 133 support, so we could at least in some limited scenarios provide a better implementation.  It would only handle the scenario where the user is doing backspace at the shell, but that would still be an improvement.

cunderw:
Submitted as #660 — off `main` at b202ea5, `public` → `open` with a note saying why, no behaviour change.

On c1b7251: that's a fair concern, and it reframes this PR — `hasText` was `return true` before that commit, so what I proposed is a revert of a deliberate change rather than a fix to an oversight. The storage-backed answer arrived with the composition rework, which is exactly where the risk of putting it back would be. Happy to wait on your testing, and to close this in favour of whatever comes out of it.

On OSC 133: agreed that beats a blanket `true` wherever the markers exist. The thing worth settling early is the fallback when no `OSC 133;A` has been seen in a session. For an SSH client that's the common case rather than the edge — the markers have to come from the *remote* shell's rc files, on a machine the app doesn't configure — so if the fallback stays `!textInputStorage.isEmpty`, the delete key still loses its repeat for most sessions. Defaulting to `true` and letting the markers *refine* it — answering `false` only when we know the cursor is at the prompt start with an empty command line — would get the improvement you're after without leaving the unconfigured case where it is today.

Glad to prototype that once you've had a look at the composition side, if it would help.

migueldeicaza:
Thanks I will do the open for now to unblock you, then I will test the scenarios, and if we still need it, then we can prototype the OSC 133 support.

cunderw:
Much appreciated!

PR #671, opened 2026-08-31
preserve Metal surface order with a detached caret

sirfergy:
A Metal surface replacement can be inserted underneath the old surface when the caret object still exists but has been removed from the view hierarchy. AppKit treats that detached ordering anchor like no anchor, placing the new surface at the bottom.

Use the caret as an anchor only while it belongs to the terminal view; otherwise retain the existing-surface fallback. Caret visibility and overlay ordering remain intact, and the change adds no public API.

## Validation

The regression exercises the real insertion helper with a detached caret and an existing surface; it fails before the fix. Adjacent cases cover an attached caret and a first surface beneath existing overlays, without requiring a GPU or window. The focused Metal suites and strict-concurrency build pass.

Upstream [baseline CI](https://github.com/migueldeicaza/SwiftTerm/actions/runs/33348635235) is already red on unrelated Linux PTY typing and Unicode segmentation failures.


migueldeicaza:

PR #673, opened 2026-08-31
add bounded copied terminal snapshots

sirfergy:
SwiftTerm keeps a view's mutable terminal private, but embedders still need input modes and a bounded copy of the viewport or history. Expose copied values captured under one terminal-lock transaction, so hosts can inspect them without retaining live terminal storage.

The snapshot remains valid after later feeds, resizes, and history trimming. Callers choose the history limit; row coordinates retain the terminal's scroll-invariant numbering.

## Tests

Covers copied-value lifetime, input modes, bounded history, complete grapheme text, row coordinates, and checked Sendable values.

## Related

Previously bundled work is now proposed independently:

- [Typed mouse responses](https://github.com/migueldeicaza/SwiftTerm/pull/681)
- [Ordered process-output consumers](https://github.com/migueldeicaza/SwiftTerm/pull/682)
- [Linux portability](https://github.com/migueldeicaza/SwiftTerm/pull/677)
- [Unicode test oracle](https://github.com/migueldeicaza/SwiftTerm/pull/678)
- [Bidi cache lifetime](https://github.com/migueldeicaza/SwiftTerm/pull/679)
- [SwiftPM shader discovery](https://github.com/migueldeicaza/SwiftTerm/pull/680)

Existing commits are preserved; the final diff contains only copied snapshots.


migueldeicaza:
The core idea is sound, but this PR contains a bunch of unrelated things, which we should review independently.

migueldeicaza:
This is a fabulous addition, I have a few suggestions/requests:

- Thread-safe does not mean safe inside callbacks. Both methods acquire the non-recursive terminal lock. A call from a delegate callback that already holds that lock will fail a precondition: This restriction is in the embedding guide; I would also put it in both public method comments so they show up during code completion.

- row.text is not ready-to-display plain text. It preserves embedded NUL characters, including wide-character continuation cells. For example, 界 can produce "界\u{0}". It removes trailing empty cells but preserves literal trailing spaces. This behavior becomes part of the public contract, The NUL is an internal detail, so I think that this should be stripped from this method.

- This is a partial snapshot. It omits row-wrap flags, hyperlinks, images, and palette values. The input snapshot also omits modes such as bracketed paste and application keypad. It is useful for inspection, but is not sufficient to reproduce the full display or implement all input encoding.

- The API is Apple-only: it is not clear if this was intended, as snapshotting seems like a useful feature to have regardless of the platform (for instance, I use HEadlessTerminal extensively for benchmarks).   It would also be useful to have public snapshot initializers also for the sake of testing.

PR #676, opened 2026-09-02
feat: blinking cursors pause as you type

freepicheep:
## Issue

Other terminals—such as Terminal.app, Ghostty, and Rio—pause a blinking cursor while the user is typing. I noticed the difference immediately in [Tecolot](https://github.com/migueldeicaza/tecolot): blinking cursors are useful, but when the cursor continues blinking during input, it can be easy to lose track of its position.

## Solution

This adds a new public `TerminalView` property, `cursorBlinkResetsOnInput`, which defaults to `true`.

When enabled, sending input makes the cursor fully visible and restarts its blink interval. Repeated input therefore keeps the cursor visible while typing, and blinking resumes after typing pauses. Setting the property to false preserves the previous behavior.

The behavior is supported consistently across:
  - macOS
  - iOS and visionOS
  - Core Animation cursor rendering
  - Metal cursor rendering

You can test the behavior in my fork of Tecolot on the [blinking-cursor](https://github.com/freepicheep/tecolot/tree/blinking-cursor) branch.

Here is a preview:
CleanShot 2026-09-02 at 10 32 23

## Testing

  - Added coverage for resetting the Metal cursor blink controller after input.

*LLM Disclosure: Codex wrote this code with Sol 5.6*

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -1348,7 +1348,14 @@ extension TerminalView {
     func configureInputSender() {
         let owner = renderOwner
         let notifyUI: @MainActor @Sendable () -> Void = { [weak self] in
-            self?.ensureCaretIsVisible()
+            guard let self else { return }
+            self.ensureCaretIsVisible()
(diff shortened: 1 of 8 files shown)


migueldeicaza:
Nice feature!

PR #679, opened 2026-09-02
prevent bidi cache reuse across line lifetimes

sirfergy:
Allocator reuse can give a new RTL paragraph the cached LTR result of a destroyed line. Keep immutable line identities alive while a paragraph-cache entry or deferred job refers to them, without retaining mutable lines or terminal buffers.

This fixes line-lifetime reuse while retaining the existing cache fast path and bounds. Font-key behavior is unchanged.

## Tests

Covers leading and wrapped-line token lifetime, release of mutable lines, equal-generation replacement, and short-lived terminal churn alongside existing bidi and snapshot behavior.

## Related

Separated from [the embedding proposal](https://github.com/migueldeicaza/SwiftTerm/pull/673) for independent review.


migueldeicaza:

PR #680, opened 2026-09-02
find SwiftPM shaders beside XCTest bundles

sirfergy:
SwiftPM can load an XCTest bundle from a separate runner, leaving the shader resources beside the bundle rather than the runner. Search that sibling location only for XCTest bundles, so Metal tests can execute without broadening application resource discovery.

Update stale test comments; legitimate device or shader unavailability remains handled as before.

## Tests

Exercises Metal concurrency, render targets, surface parity, and toggles. Local GPU coverage includes texture readback and pixel comparison, not merely an early-return path.

## Related

Separated from [the embedding proposal](https://github.com/migueldeicaza/SwiftTerm/pull/673) for independent review.


migueldeicaza:

PR #684, opened 2026-09-05
Honour the transmission medium when answering a Kitty graphics query

seiya1121:
## The problem

`handleKittyQuery` always decoded the payload as direct (`t=d`) pixel data, ignoring the transmission medium in `t=`:

```swift
private func handleKittyQuery(control: KittyGraphicsControl, base64Payload: [UInt8]) {
    guard decodeKittyPayload(control: control, base64Payload: base64Payload) != nil else {
        sendKittyError(control: control, message: "EINVAL: bad payload")
        return
    }
    ...
}
```

Clients probe for file support by sending a query whose payload is a **file path**:

```
ESC _ G i=300,a=q,t=f,f=32,s=1,v=1 ; <base64 of "/var/folders/.../frame.rgba"> ESC \
```

`decodeKittyPayload` checks that length against the `1 × 1 × 4` bytes implied by `f=32,s=1,v=1`. An 82-byte path against 4 expected bytes, so **every query for a file-based medium was answered `EINVAL: bad payload`** — even though `loadKittyPayload` already implements `t=f`, `t=t` and `t=s` correctly.

Clients read that error as "this terminal cannot do file transfers" and fall back to sending every frame **inline through the pty**, where it is base64-decoded, zlib-inflated and sliced into per-row images on the thread that feeds the terminal.

## Impact, measured

Found while embedding SwiftTerm in a SwiftUI app and running [terminal-browser](https://github.com/zenbu-labs/terminal-browser) (an Electron browser that paints into the terminal over this protocol) on a **static** page:

| | main thread | CPU |
| --- | --- | --- |
| before | **1853 of 1853** samples in `Terminal.decompressZlib` → `compression_stream_process` | 99–100%, sustained 220s with no decay |
| after | idle in `mach_msg2_trap` | **0.0%** |

After the fix the client switches to `a=T,...,t=f` and the pixels never touch the pty.

A neutral pty harness that only answers the Kitty query and counts bytes shows the same thing from the client's side, on the same static page over 30s:

| harness answers the `t=f` probe with | client uses | bytes through the pty |
| --- | --- | --- |
| `OK` | `a=T,...,t=f` (file) | **0.01 MiB** |
| `EINVAL` (what SwiftTerm did) | `a=T,...,t=d,o=z` (inline) | **1.05 MiB** |

The same page in a terminal that answers the probe correctly sits at 0.0% CPU, which is what pointed at the query path.

## The fix

Route the query through `loadKittyPayload` so it dispatches on `t=`.

That alone would be wrong, though. `t` (temporary file) and `s` (shared memory) hand ownership to the terminal, which deletes the file and unlinks the shm object after reading — but **a query transfers nothing**, so consuming the resource would destroy the very file the client was about to send. So a `consumeSource` flag is threaded through `loadKittyPayload` → `loadKittyFilePayload` → `readKittyFileData` and → `readKittySharedMemory`, and only real transmissions take ownership.

## Tests

New `KittyQueryTransmissionTests` (7 tests) covering both halves:

- a file-medium query is answered `OK`, and stores no image
- a temporary-file query leaves the file in place, while a real `a=t` transmission still deletes it (so this does not turn `t=t` into a leak)
- a missing path is still rejected
- `t=d` validation is unchanged
- an unknown medium reports `ENOTSUP` instead of being silently accepted

**Four of them fail before this change**, with `_Gi=300;EINVAL: bad payload`.

`swift test` on this branch: 739 tests, 5 failures — the same 5 `MetalRenderer` failures that are already present on the unmodified commit (732 tests, same 5). Nothing else changed.


migueldeicaza:
Great catch, thank you!  Reviewing

migueldeicaza:
Thank you for tracking this down and sending this along!

PR #689, opened 2026-09-11
Make withTerminal(_:caller:) public

arraypress:
`getTerminal()` was removed in be55ae4 when the strict-concurrency ownership boundaries landed, which is the right call — handing out the `Terminal` without the lock was the unsafe pattern.

The side effect is that embedders now have no supported way to read the terminal model at all, because `withTerminal` — the lock-holding accessor that replaced it — is internal.

Anything hosting a `TerminalView` from outside the module is stuck: checking bracketed-paste mode before pasting, reading the cursor style, pulling a line of the buffer for a prompt scrape, or issuing a reset all need the model. The remaining options are to vendor the source or to reach around the lock, and both are worse than the API.

This makes the safe path the available one. One keyword, no behaviour change. `withTerminal` already documents its contract, already takes the lock, and already records the caller for profiling, and it is used from 155 call sites inside the module, so it reads as the intended accessor rather than an implementation detail.

Happy to add a short doc note about the re-entrancy rule if you would like it spelled out for external callers.

### DIFF
--- a/Sources/SwiftTerm/Apple/AppleTerminalView.swift
+++ b/Sources/SwiftTerm/Apple/AppleTerminalView.swift
@@ -1594,7 +1594,7 @@ extension TerminalView {
     /// The closure must not call another API that synchronously acquires the
     /// terminal lock. Helpers that assume the lock is held use the `Locked`
     /// suffix and assert that contract in DEBUG builds.
-    func withTerminal<T> (_ body: (Terminal) throws -> T, caller: StaticString = #function) rethrows -> T
+    public func withTerminal<T> (_ body: (Terminal) throws -> T, caller: StaticString = #function) rethrows -> T
     {
         if ProfilingStats.enabled && Thread.isMainThread {
             ProfilingLockCallers.shared.record(caller)


migueldeicaza:
Ok, we also got #690 let me think about what is the best approach.

migueldeicaza:
Ok, the issue with this is that it undermines the entire new design of the terminal, by surfacing an object that should be carefully locked with the internal terminalLock, what this would do is make the consumer code very brittle.

You have a few options, and we should expose the options that make sense.  For example, in the case of #690 it exposes a Sendable, along with its callbacks, and the observer ran outside of the termini lock code.

For your scenario, we need to understand what it is that you need from Terminal in the first place, and surface a safe accessor to it, not just expose the internals.   Let me know what is the scenario that you needed covered, so we can add the proper accessor, we already have plenty of accessors.

migueldeicaza:
Based on your description, it seems like the only thing missing now is the status of bracketed paste, will add that to TerminalViewStateSnapshot.   Everything else you mentioned has existing safe alternatives.

- Cursor Style: terminalStateSnapshot().cursorStyle
- Pulling data: getBufferAsData(...) - but happy to add additional helpers
- Reset: softReset() or resetToInitialState()

PR #693, opened 2026-09-13
Add TerminalView.keyboardEnhancementFlags

arraypress:
## Why

SwiftTerm 2 keeps the view's `Terminal` internal, and `MigratingFrom1To2` now maps the reads a host is meant to use. One piece of state a host needs isn't on that map: whether the running application has enabled the kitty keyboard protocol.

It matters where the host sees a key before the view does. `keyDown` is `public override` rather than `open`, so an application that rebinds a chord can't override it — it installs an `NSEvent` monitor ahead of the view. My editor rebinds ⇧↩ to a newline while an agent holds the foreground, sending a backslash continuation plus Return, which is what Claude Code's own `/terminal-setup` binds in VS Code and iTerm.

When the application has switched the protocol on — Claude Code does, where the terminal offers it — SwiftTerm encodes ⇧↩ natively as Enter-with-Shift, and that monitor has to stand down or the program receives a backslash nobody asked for. `keyboardEnhancementFlags` is the only state that separates those two cases, and from a `TerminalView` host it is reachable through nothing: not the view, not `TerminalViewStateSnapshot`, not `TerminalViewDelegate`. The view's own key handling reads it for exactly this decision (`MacTerminalView` 1933, 2241, 2259; `iOSTerminalView` 2311, 2649, 2726).

## Change

- **`TerminalView.keyboardEnhancementFlags`** — a copied read, `public nonisolated`, modelled on `terminalDimensions`: `TerminalRenderOwner` takes `terminalLock` and returns the value. `KittyKeyboardFlags` is already `public` and `Sendable`, so nothing mutable is surfaced and the lock is never handed out.
- **`MigratingFrom1To2.md`** — a row in the read table and an entry under Topics.
- **`TerminalViewKeyboardEnhancementTests`** — a macOS test that pushes and pops flags through a real `TerminalView` (`CSI > 1 u` / `CSI < 1 u`), and one that sets them (`CSI = 3 ; 1 u`), both reading back through the public property.

## Testing

- `swift test` — 1423 tests in 129 suites pass, against `main` at 12ea7f5.

---

This is the accessor half of #689, which you closed — correctly. The blanket `public withTerminal` was the wrong shape, and checking my own uses against your access map moved four of six onto `pasteText`, `setCursorStyle`, `softReset()`/`resetToInitialState()` and `terminalStateSnapshot()`. `pasteText` in particular I should have found on my own; its doc comment names a dropped file's path, which was my exact case.

The only other thing keeping me on a patched copy is reading a few rows cheaply — I classify the bottom 6 rows on a timer per pane, and `terminalStateSnapshot()` copies every visible row plus a `cellWidths` array per row to answer that. A bounded row read would retire it, but it's a nice-to-have and I didn't want to bundle it in here.


migueldeicaza:
Thank you for your contributing!

PR #694, opened 2026-09-13
Cache interned attribute identifiers in front of the dictionary

ventz:
## Summary

`CellArena.intern(attribute:)` hashed the two-word `InternedAttributeKey` through `Dictionary` on every call, and it is called for almost every SGR change. On `dense_cells` that lookup (hashing plus the dynamic exclusivity check on the class property) was about 11% of the parse thread in Time Profiler.

This adds a 1,024-slot direct-mapped cache in front of the dictionary. Each entry stores both key words plus `identifier + 1` in unsafe memory, so a hit skips the hashing and the exclusivity check.

It is safe because the intern table is append-only: nothing ever removes an entry, so a cached pair is always also in the dictionary. Snapshot copies get their own empty cache.

## Headless results

`SwiftTermProfile`, 200 iterations, 5 alternating A/B pairs (order flipped each pair), against `a7b8b94`:

| Workload | Before (MiB/s) | After (MiB/s) | Change |
| --- | ---: | ---: | ---: |
| `dense_cells` | 219.7 | 270.0 | **+22.8%** |
| `medium_cells` | 138.6 | 152.4 | **+9.9%** |
| `sync_medium_cells` | 136.2 | 148.7 | **+9.2%** |
| `scrolling_fullscreen` | 775.2 | 796.8 | +2.8% |
| `unicode` | 256.2 | 262.1 | +2.3% |
| `cursor_motion` | 157.9 | 159.2 | +0.8% (noise) |
| `light_cells` | 3458.3 | 3447.9 | −0.3% (noise) |

## In a real terminal, compared with Ghostty

To check that the gain survives a real PTY, I timed a SwiftTerm-based macOS terminal ("SwiftTerm Terminal" below) before and after this change, alongside Ghostty 1.3.1 on the same M3 MacBook Pro. Each app ran the same script inside its own window at 80×24 and 12 pt, and `cat`ed a file. The clock stops on the reply to a DSR 6 cursor-position request, which a terminal can only send after it has parsed everything before it, so this measures processing and not just `cat` returning. 5 rounds, app order rotated each round, medians shown, every repetition within ±3%.

| Workload | SwiftTerm Terminal, before | SwiftTerm Terminal, with this PR | Ghostty 1.3.1 |
| --- | ---: | ---: | ---: |
| 100 MB ASCII lines | 0.327 s | 0.331 s | 1.122 s |
| 60 MB 256-color SGR text | 0.396 s | **0.313 s (+26.5%)** | 0.789 s |
| 40 MB CJK and emoji | 0.255 s | 0.254 s | 0.322 s |
| 5,000 full-screen redraw frames | 0.150 s | 0.149 s | 0.740 s |

With this change SwiftTerm Terminal is 2.5× faster than Ghostty on colored output, 3.4× on ASCII, 1.3× on CJK/emoji and 5× on full-screen redraws. ASCII does not move because it is already at the macOS PTY ceiling described in `PERFORMANCE.md` (~290 MiB/s). These are throughput numbers only; input latency was not measured.

## Testing

- `swift test`: 1,357 tests. The only failure is a Kitty shared-memory test (`testKittySharedMemoryLoad`), which fails the same way on the untouched base in this environment.
- Also tried and rejected: walking `EscapeSequenceParser.parseBorrowed` by index instead of `extracting(droppingFirst:)`. It lost 3–6% on every engine-bound workload, even with `data[unchecked:]`.


migueldeicaza:
Whoa, this is amazing!   Let me review this.

migueldeicaza:
This looks fabulous, I do wonder if we can use a more modern API that does not rely on Unsafe allocations.

Let me research, these numbers are fabulous.

migueldeicaza:
Ok added a change that does not use unsafe code.

I tested three scenarios, I wish I could have used InlineArray,  but it is only available on MacOS 26 and there is no easy way of using it dynamically:

Common macOS 26 comparison, five interleaved fixed-work runs, p0 throughput versus no cache:

| Storage | `dense_cells` | `medium_cells` | `sync_medium_cells` |
|---|---:|---:|---:|
| `UnsafeMutablePointer` | +31.3% | +12.8% | +12.7% |
| `Array` | +28.0% | +12.3% | +11.2% |
| `InlineArray` | +32.0% | +12.8% | +12.4% |

ventz:
That's great!  - thank you!

migueldeicaza:
I had to do a follow up change, as we regressed some 2% on Unicode, but it is all good now.

ventz:
Nice find!

PR #696, opened 2026-09-15
Notify hosts of in-place row shifts and count buffer resets

henrymodisett:
## What

Hosts that track their own absolute row anchors (my terminal's prompt markers and read cursors) have no observable signal when rows shift in place: `selectionsAdjustForInPlaceScroll` translates only attached selections. A top-anchored margin scrolled against full scrollback trims the front row while rows below the margin stay put, so absolute anchors below the margin resolve one row too high per scroll. Separately, a hard reset on same-geometry buffers changes no counter a polled buffer fingerprint can see (`linesTop` 0→0, same dimensions) — and a reset from the normal buffer returns before `selectionsClear()`, so attached selections go stale the same way.

## Change

- `Terminal.onRowsShiftedInPlace` closure, invoked from the single `selectionsAdjustForInPlaceScroll` choke point with the same region and direction plus the post-shift trim count, so hosts apply the same translation to their own anchors. A closure rather than a delegate method: hosts embedding the SwiftTerm views inherit the delegate conformance, and protocol witnesses cannot be specialized from a subclass in another module. Nil by default; zero cost when unset.
- `Buffer.resetEpoch`: counts wholesale content replacements (bumped in `clear()`, carried across `resetNormalBuffer()` rewiring). Fresh buffers start at zero; equal epochs with equal coordinates stay comparable.

Both are additive: no existing behavior changes, all existing tests pass (one pre-existing failure in `SelectionScrollTests/testFullScreenScrollLeavesSelectionAlone` reproduces on the base commit).

## Tests

New `Tests/SwiftTermTests/RowShiftNotificationTests.swift`: hook args pinned for partial-margin trim, full-width scroll, and pre-full splice (region, direction, trim count), plus epoch 0 → 1 → 2 across hard resets with trims leaving it alone.


henrymodisett:
Updated the proposal after review of the consumer side: the shift mirror now carries `linesTopDelta` (how much of the trim count each shift consumed) so hosts translate with one law — pre-shift window membership, motion net of the consumption, post-shift landing, outside anchors absorbing the base share — instead of inferring growth from a shadow sighting. This also fixes a real missed mirror: `Terminal.clearScrollback` trimmed without notifying. The branch head is now 990da07 (tag `vesper-pin-v1.20.0-5` on the fork).

migueldeicaza:
I think you accidentally pushed a bunch of commits for something else, and now the patch contains lot of unrelated changes to this PR.   I suggest we start with a fresh one.

henrymodisett:
You were right, thanks for the steer — reopened as #708: the same proposal on a branch that carries only it, rebased onto current main.

PR #698, opened 2026-09-15
Report progress state to the host

onury:
## Why

A host that embeds a terminal usually wants to know what the session is doing — show a spinner next to its tab while a command runs, hold queued input back until it finishes, badge a window that is still working. The view already knows: it handles the OSC 9;4 progress reports and draws the bar from them. It just keeps that to itself, so an embedding application has no way to read it short of scraping the screen.

## What changed

Two defaulted methods on `TerminalViewDelegate`, on macOS and iOS:

- `progressReport(source:report:)` — called with every OSC 9;4 report the view handles, right after it applies it to the bar. The 15 second silence timer now runs through the same path and sends a synthetic `.remove`, so `set` and `remove` always arrive in pairs and a host that mirrors the state never sticks on busy.
- `notification(source:title:body:)` — called for OSC 777 notifications. The terminal already parsed those and the view discarded them; whether one becomes a banner, a badge, or nothing at all is the host's policy, so this only hands it over. SwiftTerm still posts nothing itself.

`LocalProcessTerminalView` is its own `TerminalViewDelegate`, so a host that uses it cannot answer either callback — the conformance is satisfied by the protocol's default implementation, and a subclass method does not replace it. Both are therefore also on `LocalProcessTerminalViewDelegate`, defaulted, and forwarded by the view (`open`, so a subclass can answer them directly instead) — the same shape as the Kitty clipboard hooks.

The progress bar is untouched — same drawing, same timer, same states. Both methods have empty default implementations in the shared `TerminalViewDelegate` extension, so no existing conformer breaks.

Documented under a new "Session Signals" section in `Customization.md`.

## Tests

`Tests/SwiftTermTests/SessionSignalTests.swift` covers the determinate, indeterminate and removal reports, the synthetic removal the silence timer sends, a malformed report that must not be forwarded, OSC 777 (including a body with a semicolon in it), that the bar still hides and shows exactly as before, the forwarding through `LocalProcessTerminalView`, and a delegate implementing neither new method — which is a compile-time check that both stay optional.

`swift build` and `swift test` pass (`testKittySharedMemoryLoad` fails on `main` here too, unrelated); the iOS build was checked with `xcodebuild -destination 'generic/platform=iOS'`.


migueldeicaza:
This makes sense, I'll check it out.

migueldeicaza:
This fixes timer expiration, but updateUiClosed() still calls clearProgressReport() directly on macOS and iOS. A delegate that received an active report can therefore remain busy after shutdown. Route an active shutdown through the reporting path, or narrow the documented guarantee. The “always come in pairs” wording is also too strong because terminalDelegate can change while progress is active.

Also, it conflicts with the other PR.

onury:
You are right on both counts. `updateUiClosed()` now goes through the reporting path when a report is live, so the delegate receives a `remove` before the view stops; with nothing live it clears as before, and a second close says nothing. The wording is narrowed to what the view can promise: a `set` or `indeterminate` is followed by a `remove` from the application, the silence timer, or the view closing, as long as the same delegate stays attached. A delegate attached mid-report is not caught up, and one detached before the removal never hears it. Three tests cover the close cases beside the existing expiration one.

#704 is rebased onto this branch, so the conflict is gone; it should land after this one.

PR #700, opened 2026-09-16
macOS: report focus on window activation; keep Control in Meta chords on iOS

ardhitama:
These are two fixes we carry as local patches in the Ardvis fork (a terminal app built on SwiftTerm), submitted to see whether you want them upstream. Happy to split, rewrite or drop either one.

## macOS: focus reporting does not follow window activation

DECSET 1004 reports were driven only by first-responder changes. AppKit keeps a window's first responder when the window loses key status, so a terminal in a background window kept reporting itself as focused and never reported the loss — switching to another application left the program running in the terminal believing it still had focus. `hasFocus` already consulted `window?.isKeyWindow`, but nothing recomputed the terminal's focus when activation changed.

Focus is now recomputed in one place, `updateTerminalFocus()`, called from `viewDidMoveToWindow`, from the `didBecomeKey`/`didResignKey` notifications for the view's own window, and from `becomeFirstResponder`/`resignFirstResponder`. The notification handlers now ignore events for other windows, instead of updating the caret and marking the view dirty for any window on the system. Pending kitty key releases are cleared when focus is lost, since their key-up events never arrive.

`Tests/SwiftTermTests/WindowFocusReportTests.swift` drives a window whose key status the test controls. It fails on the current `main` at both activation assertions (`didResignKey` and `didBecomeKey` produce no report) and passes with the change.

## iOS: Meta+Control chords lose the Control

In the legacy key path, when `optionAsMetaKey` (or a Meta modifier key) is held together with Control, the input went out as `ESC` plus the unmodified character. The application received Meta-`<key>` and the Control was dropped: Ctrl+Option+C reached it as `ESC c` (RIS) rather than `ESC 0x03`. The enhanced keyboard encoder already prefixes ESC to the Control-transformed character; the legacy path now does the same.

## Verification

- `swift test` on macOS: 1424 tests in 130 suites pass.
- `swift build --triple arm64-apple-ios17.0-simulator`: builds.
- The macOS test fails without the first change and passes with it.
- The iOS change has no automated test — `swift test` cannot drive the UIKit input path. It is exercised by hand in our app.


migueldeicaza:
Thank you, let me take a look!

Sources/SwiftTerm/iOS/iOSTerminalView.swift
```diff
@@ -3300,7 +3300,13 @@ open class TerminalView: UIScrollView, UITextInputTraits, UIKeyInput, UIScrollVi
                 if key.modifierFlags.contains ([.alternate, .command]) && key.charactersIgnoringModifiers == "o" {
                     optionAsMetaKey.toggle()
                 } else if (key.modifierFlags.contains (.alternate) && optionAsMetaKey) || metaModifier {
-                    data = .text("\u{1b}\(key.charactersIgnoringModifiers)")
+                    // Meta prefixes the Control-transformed character for a
+                    // combined chord, just as in the enhanced keyboard encoder.
+                    if key.modifierFlags.contains(.control) {
+                        data = .bytes([0x1b] + applyControlToEventCharacters(key.charactersIgnoringModifiers))
```
migueldeicaza: Small problem here: applyControlToEventCharacters can return an empty array. For combinations such as Control+Option+1, this change sends only ESC and consumes the key. Previously it sent ESC plus the original character. Use the control bytes only when they are nonempty; otherwise keep the existing Meta text fallback.

migueldeicaza (commented):
Looks pretty good, just one small bug,

Issue #701, opened 2026-09-16
Implicit link regex joins adjacent paths and swallows prose after undotted paths


mcclowes:
## Summary

The implicit link regex (`ghosttyImplicitLinkRegex` in `Terminal.swift`) matches far more than the path in two common cases. Tested against 1.18.0 (`7691f85`) by compiling the regex exactly as the file builds it.

### 1. A space-separated path absorbs the next path

Ghostty's space segments refuse to continue into a word that starts a new path. Its `src/config/url.zig` has:

```
(?:(?<!:) (?!\w+:\/\/)(?!\.{0,2}\/)(?!~\/)[\w\-.~:\/?#@!$&*+;=%]*[\/.])*
```

SwiftTerm's `dottedPathSpaceSegments` and `anyPathSpaceSegments` drop `(?!\.{0,2}\/)(?!~\/)`, so two paths on one line become one link:

| Input | SwiftTerm today | With Ghostty's lookaheads |
| --- | --- | --- |
| `cp /a/b ./c/d` | `[/a/b ./c/d]` | `[/a/b]` `[./c/d]` |
| `mv ~/x/y ~/z/w` | `[~/x/y ~/z/w]` | `[~/x/y]` `[~/z/w]` |
| `cd /a/b; ls /c/d` | `[/a/b; ls /c/d]` | `[/a/b; ls]` `[/c/d]` |
| `/tmp/test folder/file.txt` | `[/tmp/test folder/file.txt]` | unchanged |

**Suggested fix:** add `(?!\.{0,2}\/)(?!~\/)` after `(?!\w+:\/\/)` in both space-segment patterns, matching upstream Ghostty.

### 2. A path without a dot swallows the rest of the line

When the first path run has no `.`, `anyPathSpaceSegments` accepts any following word, so prose after a path is underlined and returned as the link:

```
for /settings/byok#routers when a router is connected but has no saved models. The chat
→ [/settings/byok#routers when a router is connected but has no saved models. The chat]
```

This matches Ghostty's behavior (its tests expect `./space middle` to match whole), so it's a deliberate tradeoff for paths with spaces rather than a porting bug. It shows up constantly in output from coding agents like Claude Code and Codex, which print routes and paths inline in prose.

Some options:

- Only continue past a space when the continued word itself looks like a path segment (contains `/` or `.`), as the dotted branch already requires. This loses undotted names like `/Users/me/My Documents`.
[...]

PR #702, opened 2026-09-16
Stop implicit path links from absorbing the next path

mcclowes:
Part of #701.

The implicit link regex's space segments were missing Ghostty's `(?!\.{0,2}\/)(?!~\/)` lookaheads (see `src/config/url.zig` upstream), so a path followed by another path on the same line matched as a single link:

| Input | Before | After |
| --- | --- | --- |
| `cp /a/b ./c/d` | `/a/b ./c/d` | `/a/b`, `./c/d` |
| `mv ~/x/y ~/z/w` | `~/x/y ~/z/w` | `~/x/y`, `~/z/w` |
| `diff /tmp/a.txt /tmp/b.txt` | `/tmp/a.txt /tmp/b.txt` | `/tmp/a.txt`, `/tmp/b.txt` |
| `/tmp/test folder/file.txt` | unchanged | unchanged |

Paths with interior spaces still match, since a continuation is only refused when it starts with `/`, `./`, `../`, or `~/`.

This doesn't address the second half of #701, where an undotted path absorbs following prose. That matches Ghostty's behavior, so it seemed like a design question rather than part of this fix.

## Testing

- Added `testAdjacentPathsAreSeparateMatches`, which fails before the change.
- `GhosttyImplicitLinkDetectionTests`, `LinkLookupTests`, and the Ghostty fuzz suites pass.
- In the full `swift test` run, `testKittySharedMemoryLoad` fails on `main` without this change too, and `endCallbackDoesNotOverlapFinalParse` timed out under load but passes when run alone.


migueldeicaza:
Thank you for your contribution. merging!

PR #703, opened 2026-09-17
Forward the bell through LocalProcessTerminalView to its process delegate

onury:
`LocalProcessTerminalView` is its own `TerminalViewDelegate`, so a host that embeds it never hears the bell: `TerminalViewDelegate`'s default `bell(source:)` beeps and that is the end of it, and a method on a subclass outside the package does not replace a protocol-extension witness.

This adds `bell(source:)` to `LocalProcessTerminalViewDelegate`, with a default that beeps exactly as before, and an `open` override on the view that forwards to `processDelegate`. Existing hosts see no change; a host that wants to show the bell some other way (a badge, a notification when its window is not key) implements the delegate method.

Same shape as the existing `setTerminalTitle` / `hostCurrentDirectoryUpdate` forwards. Builds clean on macOS.


Sources/SwiftTerm/Mac/MacLocalTerminalView.swift
```diff
@@ -350,6 +359,13 @@ open class LocalProcessTerminalView: TerminalView, TerminalViewDelegate {
         processDelegate?.hostCurrentDirectoryUpdate(source: source, directory: directory)
     }
 
+    /// The view is its own `terminalDelegate`, so the bell would otherwise
+    /// stop at the protocol's default beep with no way for the host to hear
+    /// it. `open`, so a subclass can also answer it directly.
+    open func bell(source: TerminalView) {
+        processDelegate?.bell(source: source)
+    }
```
migueldeicaza, line 375: ProcessDelegate?.bell(...) does nothing when processDelegate is nil. Today, the TerminalViewDelegate default calls NSSound.beep(). The PR therefore does not preserve existing behavior as claimed. It needs an explicit fallback:

migueldeicaza (changes requested):
In Tecolot, you can see how I solved this problem, which is to subclass the terminal, but I appreciate that this might be useful for quick customizations.   

Please check the review below.

onury:
You are right, and the PR's claim was wrong. `LocalProcessTerminalView` conforms to `TerminalViewDelegate`, so before this change the bell landed on that protocol's own default and beeped. Overriding it to forward replaced that beep with silence for anyone who never set a `processDelegate`.

Fixed in the second commit — the override falls back to the beep when there is no delegate:

```swift
open func bell(source: TerminalView) {
    guard let processDelegate else {
        NSSound.beep()
        return
    }
    processDelegate.bell(source: source)
}
```

`LocalProcessTerminalViewDelegate` already defaults `bell(source:)` to `NSSound.beep()`, so the other gap — a host that sets a delegate but does not implement the method — was covered. The nil case is the one that regressed.

The sibling forwards on this view, `progressReport` and `notification`, have no beep to lose: their `TerminalViewDelegate` defaults are empty, so forwarding to nothing is what already happened.

On subclassing: that is the right answer when you own the view, and it is why every forward here is `open` rather than `final` — a subclass can still answer directly and ignore the delegate. The delegate path is for a host that already holds a `LocalProcessTerminalViewDelegate` for the process lifecycle and wants the bell to arrive at the same place, without a subclass existing only to move one call.

PR #704, opened 2026-09-17
Let the host turn the built-in progress bar off while still receiving the reports

onury:
A host that draws progress in its own chrome (a bar in a toolbar or a footer, say) has no way today to stop `TerminalView` drawing the OSC 9;4 bar inside the terminal as well — the two end up on screen together.

This adds `showsProgressBar` (default `true`, so nothing changes for anyone) on both the macOS and iOS views. While it is `false` no report brings the bar back, and setting it to `true` again restores it — a report still live comes back with it. The reports themselves still reach an observer registered with `observeOscEvents(_:)`, so the host keeps what it draws from.

Mechanically: the last live report is kept beside the bar view and one `syncProgressBar()` is the only place a report or the flag reaches the view, so the two cannot disagree. The 15 s silence timer is untouched. A docs section and six tests (default draws; off keeps every report off screen; the observer still hears them; off takes a live bar down; on restores a live report; nothing comes back after the program finished while it was off). Builds and tests clean on macOS; the iOS scheme builds.


migueldeicaza:
This uses the wrong observation API for a custom bar.

 That observer receives raw OSC input, including malformed reports, but it does not receive the synthetic removal from the 15-second timer. Thus, a custom bar can remain busy after SwiftTerm clears its internal state. 

We probably should merge #698 first, then document and use TerminalViewDelegate.progressReport. Add a test for showsProgressBar = false, progress expiration, and delegate removal.   That said, 698 had a separate issue.

onury:
Agreed, the observer was the wrong door. The branch now sits on #698: the host draws from `TerminalViewDelegate.progressReport`, which carries the synthetic `remove` on expiration and on close, and the docs say so instead of pointing at `observeOscEvents`. Tests added for the three cases you named: bar off with reports still reaching the delegate, expiration with the bar off, and the delegate dropped mid-report followed by expiration and close.

PR #705, opened 2026-09-18
Let the host make Option-drag bypass mouse reporting, the way Shift does

onury:
When a program turns mouse reporting on, every drag goes to it and the view's own selection is out of reach except through Shift. Every other macOS terminal also honors Option for this — it is what programs mean when they print "option+click to select" on macOS (Claude Code does, for one).

This adds `optionBypassesMouseReporting` on the macOS view, `false` by default so nothing changes for anyone. On, an Option-modified mouse event skips reporting and works the selection, exactly as Shift already does through the same predicate. One line in the predicate, one documented property. Builds clean on macOS.


migueldeicaza:
I am generally fine with this setting, but I tried midnight command on a few terminals and I would love to know what I should be testing for:

* Terminal.app: shift-drag does nothing, nor does option-drag
* Ghostty: shift-drag selects, option-drag does nothing
* SwiftTerm: shift-drag selects, option-drag does nothing

The operation I performed was just dragging the selection on the terminal.

onury:
The setting is off by default, so all three of those runs were comparing terminals with it disabled — SwiftTerm included. Nothing changes until the host opts in:

```swift
terminalView.optionBypassesMouseReporting = true
```

What to test, and what to expect:

1. Run a program that turns mouse reporting **on** — `mc` qualifies, so does anything full-screen with mouse tracking.
2. With the flag off (today's behaviour), Option-drag sends mouse reports to the program. The program decides what that means, which for most is nothing visible.
3. With the flag on, Option-drag never reaches the program. The view does its own selection, exactly as Shift-drag does now, and ⌘C copies it.

So the observable difference is only between those last two, on one terminal, with one line changed.

The case that motivated it: a full-screen program that takes the mouse and paints its own selection. The host's selection colour is then unreachable — the user drags, the program highlights in its own colours, and nothing the host configured applies. Option-drag gives the selection back to the view without the program having to cooperate. iTerm2 is the closest reference: holding Option there temporarily suspends mouse reporting for the same reason.

Shift already does this in SwiftTerm, gated on `mouseShiftCapture`. This is the same bypass on a second modifier, off unless asked for, so nothing changes for an existing host.

PR #707, opened 2026-09-19
Fix Insert/Delete Line corrupting the screen in margin mode

jhancock1975:
### The bug

In margin mode (DECLRMM, `CSI ? 69 h`), `cmdInsertLines` and `cmdDeleteLines` copy rows starting at the **cursor row** but size the copy with the scroll region's **full height**:

```swift
let rowCount = buffer.scrollBottom - buffer.scrollTop
let src = buffer.lines [row+i+1]        // row = buffer.y + buffer.yBase
```

The number of rows to move is the distance from the cursor to the bottom of the region, not the region's height. With the cursor low in the region the loop indexes past the last line, and `CircularList`'s subscript wraps modulo its backing array, so the tail of the copy overwrites rows at the **top** of the screen. The same over-long range is handed to the `repairWideCellsForColumnRestrictedShift` loop just above it.

`cmdScrollUp` and `cmdScrollDown` use the same idiom but start at `buffer.scrollTop`, where the region height *is* the right count — so they are correct and untouched here.

Delete Line's margin branch was also missing the vertical-region guard that its non-margin branch has (`buffer.y >= scrollTop && buffer.y <= scrollBottom`), so it acted even with the cursor outside the region. Insert Line already guards this at the top of the function.

### How to see it

tmux enables DECLRMM at startup and never turns it off, so anything full-screen inside tmux can hit this. A minimal reproduction — the only difference between the two streams is `ESC [ ? 69 h`:

```
ESC[?1049h ESC[2J ESC[H          (fill 24 rows with ROW01..ROW24)
ESC[?69h                          <-- with this line, the screen is destroyed
ESC[1;22r ESC[20;1H ESC[2M
```

Without margin mode: rows 1-19 untouched, `ROW22` moves up to row 20, rows 21-22 blank, `ROW23`/`ROW24` untouched. With it: the whole top of the screen shifts and `ROW01` reappears on the last row.

In practice this came from a user report of the emacs mode line being garbled inside tmux after saving or paging — the menu bar is destroyed and the mode line lands in the wrong row.

### Verification

I drove this emulator against a real pty running `tmux` + `emacs` and used `tmux capture-pane` as the oracle, so any row where the emulator disagrees with what tmux believes it drew is a defect. **14 rows disagreed before this change, 0 after**, deterministically across runs.

Added `Tests/SwiftTermTests/MarginModeLineEditTests.swift` with four tests: DL and IL leave rows above the cursor alone, the margin path matches the non-margin path at full width, and DL outside the scroll region is ignored. Three of the four fail without the fix.

Full suite: two unrelated failures on this machine — `testKittySharedMemoryLoad` (also fails on an unmodified checkout) and `cursorBlinkControllerOwnsTimerOnMainActor` (flaky under parallel load, passes in isolation on both).


migueldeicaza:

I can confirm the bug and the patch are fine.

